PR #13, opened 2018-01-19
(PDK-506) Resource API-based provider template

DavidS:


DavidS:
[v0.2.0](https://rubygems.org/gems/puppet-resource_api/versions/0.2.0) was just released, which should address that particular issue.

PR #14, opened 2018-01-24
(maint) Adds spec_helper_local support to spec_helper.rb

bmjen:


DavidS:
another possibility would be to recommend moving this content to `spec/support`. See e.g. https://github.com/puppetlabs/puppetlabs-azure/tree/master/spec/support (through https://github.com/puppetlabs/puppetlabs-azure/blob/d085dbf4c0345f687fcd1dd586a72a91d7940285/spec/spec_helper_acceptance.rb#L26-L27 )

But also http://ryanbigg.com/2013/02/about-spec-support , which might not be applicable, because I would not expect tons of content in there.

HelenCampbell:
We are currently adding the require line into .sync.yml for modules with spec_helper_local present, I'd back this change 👍

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -1,5 +1,12 @@
 require 'puppetlabs_spec_helper/module_spec_helper'
 require 'rspec-puppet-facts'
+
+begin
+  require 'spec_helper_local'
+rescue LoadError => loaderror
+  puts "Could not require spec_helper_local: #{loaderror.message}"
```
rodjek: `s/puts/warn/` so that this is printed to stderr and can be disabled with other warnings. Also, I'd recommend guarding this to only attempt the require if `spec/spec_helper_local.rb` exists, otherwise it'll print an unnecessary error for people who don't need/use a spec_helper_local

bmjen: @rodjek updated with suggestions.

rodjek:
modified to `s/__File__/__FILE__/`

PR #16, opened 2018-01-25
Removing openssl issue from appveyor config as is no longer required

HelenCampbell:
(https://bugs.ruby-lang.org/issues/11033)
This workaround is no longer required and should be removed.
The PDK is reverting the removal of this which was rolled out by modulesync as can be seen in this PR:
https://github.com/puppetlabs/puppetlabs-concat/commit/1587974bd7e1330a6ec329a54ccb97647f302c08


glennsarti:
This change looks ok.   @iristyle any thoughts?

Iristyle:
Yup, this should be fine ... same as the removal at https://github.com/puppetlabs/modulesync_configs/pull/175

PR #20, opened 2018-01-31
(PDK-776) Refactors ignore files for git and pdk, removes pmtignore.

bmjen:


HelenCampbell:
You also (accidentally?) removed .DS_Store :)
I second Davids point about people not being able to add to the ignore list without replacing content - During our converts we are hitting in to something similar with rubocop defaults when trying to add another require and it's frustrating.

PR #23, opened 2018-02-07
Succeed with unsupported rubocop versions

DavidS:
This fixes https://github.com/puppetlabs/pdk/issues/420


DavidS:
h/t @igalic.

PR #28, opened 2018-02-21
Fix geppetto metadata to work with newer versions of eclipse

hdeheer:
nature name is org.eclipse.. instead of com.eclipse...


DavidS:
@hdeheer are you still using geppetto? Is this what's necessary today?

hdeheer:
@DavidS Correct, still using Eclipse with the Geppetto plugin. Will probably switch to RubyMine someday.

DavidS:
:+1:

for a "lighter-weight" experience you might also want to look into the [puppet plugin for VSCode](https://marketplace.visualstudio.com/items?itemName=jpogran.puppet-vscode).

PR #31, opened 2018-02-26
Removing duplicate log value in gitignore

HelenCampbell:


HelenCampbell:
Cheeky self-merge because this is pretty straight forward.

PR #32, opened 2018-03-03
Make deploy optional in .travis.yml

ardrigh:
Deployments from CI to the Puppet Forge should be configured as opt-in. A lot of internal or private modules will not want these lines is the CI configuration.

Also add the option to specify a valid username along with the secure password.


edestecd:
@ardrigh do you need to add some defaults to: https://github.com/puppetlabs/pdk-templates/blob/master/config_defaults.yml

bmjen:
@ardrigh  I like these changes, but I think the defaulting needs to be to `true` and the username to `puppet`, so that we can maintain compatibility for those PDK users that currently rely on this setting.

ardrigh:
@bmjen I added the defaults to `config_defaults.yml` and did some brief testing.

A new module is created with the deploy section

If a user creates `.sync.yml`:
- `pdk update` will add the section if `deploy: true`, with the default `user: puppet`
- `pdk update` will add the section if `deploy: true`, with user name supplied e.g. `user: foo`

Is that suitable, or would you prefer to have the defaults in the template?

bmjen:
Thanks @ardrigh ! That's perfect. I just wanted to make sure previous behavior is maintained, which is... a PDK user that created a new module with the older templates, does not need to add a `.sync.yml` update in order to keep their module configuration the same.

PR #38, opened 2018-03-13
(MODULES-6828) Add blacksmith to Gemfile and Rakefile

HelenCampbell:


config_defaults.yml
```diff
@@ -524,6 +524,8 @@ Gemfile:
           - mswin
           - mingw
           - x64_mingw
+      - gem: puppet-blacksmith
+        version: '~> 3.4'
```
bmjen: This has to have a guard that will only install on ruby platforms or it will break windows.

HelenCampbell: Fixed :)

PR #42, opened 2018-03-30
(PDK-706) Remove vulnerable puppet3 support dependencies

### LINKED ISSUE #24: Remove legacy version checks from Gemfiles
The new modules generated by pdk 1.3 creates a Gemfile which includes version tests for old 3.x versions of Puppet.

That not only adds a number of lines to the file, it's also slightly confusing reading it when the PDK documentation says it only supports Puppet 4 onwards.

I see there was a [PR in the old repo](https://github.com/puppetlabs/pdk-module-template/pull/62) to try and remove some of these lines.

Is this change going to be applied here to the [current template](https://github.com/puppetlabs/pdk-templates/blob/master/moduleroot/Gemfile.erb)?

bmjen:
This replaces the old PR at https://github.com/puppetlabs/pdk-module-template/pull/62 and removes the legacy puppet checks and pins.

Closes https://github.com/puppetlabs/pdk-module-template/pull/62
Fixes #24

### DIFF
--- a/moduleroot/Gemfile.erb
+++ b/moduleroot/Gemfile.erb
@@ -100,73 +100,24 @@ puppet_type = gem_type(puppet_version)
 facter_version = ENV['FACTER_GEM_VERSION']
 hiera_version = ENV['HIERA_GEM_VERSION']
 
-def puppet_older_than?(version)
-  puppet_version = ENV['PUPPET_GEM_VERSION']
-  !puppet_version.nil? &&
-    Gem::Version.correct?(puppet_version) &&
-    Gem::Requirement.new("< #{version}").satisfied_by?(Gem::Version.new(puppet_version.dup))
-end
-
 gems = {}
 
 gems['puppet'] = location_for(puppet_version)
 
 # If facter or hiera versions have been specified via the environment
-# variables, use those versions. If not, and if the puppet version is < 3.5.0,
-# use known good versions of both for puppet < 3.5.0.
-if facter_version
-  gems['facter'] = location_for(facter_version)
-elsif puppet_type == :gem && puppet_older_than?('3.5.0')
-  gems['facter'] = ['>= 1.6.11', '<= 1.7.5', require: false]
-end
+# variables
 
-if hiera_version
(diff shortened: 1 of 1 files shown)


bmjen:
@Iristyle This also includes your suggestions I think.

moduleroot/Gemfile.erb
```diff
@@ -100,73 +100,23 @@ puppet_type = gem_type(puppet_version)
...
-  gems['windows-pr'] =          ['1.2.3', require: false]
-elsif Gem.win_platform?
+if Gem.win_platform? && puppet_version =~ %r{^(file:///|git://)}
   # If we're using a Puppet gem on Windows which handles its own win32-xxx gem
   # dependencies (>= 3.5.0), set the maximum versions (see PUP-6445).
   gems['win32-dir'] =      ['<= 0.4.9', require: false]
   gems['win32-eventlog'] = ['<= 0.6.5', require: false]
   gems['win32-process'] =  ['<= 0.7.5', require: false]
   gems['win32-security'] = ['<= 0.2.5', require: false]
-  gems['win32-service'] =  ['<= 0.8.8', require: false]
```
Iristyle: Ok - so if the intent is to remove the win specific deps from the ubergems... and instead use the Gemfile as a fallback (which is what we used to do before the ubergems), you'll still need to keep `win32-service` here since Puppet depends on it and it's not in the gemspec:

https://rubygems.org/gems/puppet/versions/5.5.0-x64-mingw32

Iristyle: Also ... what about FFI and minitar?

Facter 2 depends on FFI (and would therefore pull it in), but Facter 3 is on its way... and it does not depend on FFI.

Iristyle:
See also commentary on https://github.com/puppetlabs/modulesync_configs/pull/163

moduleroot/Gemfile.erb
bmjen: Are you suggesting adding a similar conditional block to bring in an ffi dep if FACTER_GEM_VERSION points to a file or git repo?

DavidS:
:+1:

PR #44, opened 2018-03-30
Proper matrix testing of ruby<->puppet

cdenneen:


npwalker:
I think this is essentially a dead PR now after #71 went in.

PR #48, opened 2018-04-06
(PDK-911) fix ensure symbol/string handling

DavidS:
Requires to be synchronised with https://github.com/puppetlabs/puppet-resource_api/pull/55


DavidS:
This can be merged now.

PR #49, opened 2018-04-10
(PDK-916) Avoid the PSH deprecation when not specifying mock_with

DavidS:


DavidS:
This is related to https://github.com/puppetlabs/puppetlabs_spec_helper/pull/239

DavidS:
Notably this only seems to work with current puppetlabs_spec_helper master, but not with the released 2.6.2 :-(

PR #57, opened 2018-04-22
(PDK-944) Enforce WordArray brackets style

jearls:
See https://tickets.puppetlabs.com/browse/PDK-944


dhollinger:
This looks like a good solution to me

jearls:
this looks fine to me as well.  my only concern is changing the enforced style rather than un-enforcing it.  in SemVer, this would be the sort of thing that requires a major version change, so i was trying to avoid that myself.

DavidS:
@bmjen I think this can move forwards.

jearls:
rebased off current master

jearls:
Thanks!

PR #65, opened 2018-05-01
Add yaml header to make yamllint happy

wmuizelaar:


npwalker:
@bmjen this seems reasonable to me.

PR #68, opened 2018-05-01
(PDK-964) Use gitlab ci caching

npwalker:


npwalker:
@dhollinger thoughts?

npwalker:
@DavidS @bmjen thoughts?

dhollinger:
@npwalker how does caching work when bundler is run within separate containers for each job?

npwalker:
@dhollinger I think the cache works similar to how Travis does caching.  

https://docs.gitlab.com/ee/ci/caching/#caching-ruby-dependencies

It's going to make a tar of all the installed gems the first time and then on subsequent test runs it will just untar the gems instead of installing them via `bundle install`.  `bundle install` after the cache is extracted should correct any differences.  

There are options to make caching more granular, per-branch, per-pr, etc...

https://docs.gitlab.com/ee/ci/caching/#sharing-caches-across-the-same-branch

dhollinger:
Ok. Looks good to me then

PR #69, opened 2018-05-01
Default to parallel specs in CI

npwalker:
Prior to this commit, we ran `rake spec` instead of
`rake parallel_spec`.  Since every travis, appveyor, and
gitlab ci all provide at least 2 cores running in parallel
can reduce runtimes.

After this commit, we default to running parallel_spec.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -52,8 +52,8 @@
     - env: CHECK="syntax lint"
     - env: CHECK=metadata_lint
     - env: CHECK=release_checks
-    - env: CHECK=spec
-    - env: PUPPET_GEM_VERSION="~> 4.0" CHECK=spec
+    - env: CHECK=parallel_spec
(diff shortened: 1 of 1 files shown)


npwalker:
@DavidS @bmjen thoughts?

config_defaults.yml
```diff
@@ -533,6 +533,8 @@ Gemfile:
         require: false
         platforms:
           - ruby
+      - gem: 'parallel_tests'
```
DavidS: This is already being pulled in by the default gemset. Did you have a situation where this didn't happen?

npwalker: I'm pretty sure I tried to run parallel_spec and it gave me an error without it added but I can test again when I rebase this.

npwalker: Yea, looks like parallel_tests is there by default so I removed that addition to the Gemfile.

PR #70, opened 2018-05-01
(PDK-963) Only run rubocop, syntax, metadata on first ruby version

npwalker:
Prior to this syntax checks on ruby 2.1.9 fail due to a
deprecation warning.

### DIFF
--- a/moduleroot/.gitlab-ci.yml.erb
+++ b/moduleroot/.gitlab-ci.yml.erb
@@ -13,7 +13,7 @@ before_script:
   - bundle -v
   - bundle install <%= @configs['bundler_args'] %>
 
-<% @configs['ruby_versions'].each do |ruby_version| -%>
+<% ruby_version = @configs['ruby_versions'].first -%>
 rubocop-<%= ruby_version %>:
   stage: test_<%= ruby_version %>
   image: ruby:<%= ruby_version %>
@@ -32,6 +32,7 @@ metadata-<%= ruby_version %>:
   script:
     - bundle exec rake metadata_lint
 
+<% @configs['ruby_versions'].each do |ruby_version| -%>
 rspec-puppet-<%= ruby_version %>:
   stage: test_<%= ruby_version %>
   image: ruby:<%= ruby_version %>


cdenneen:
@dhollinger this was something we discussed in slack with @rnelson0... Should we be restricting some of these tests for "speed" and "resources" to a single `ruby_version`?

@DavidS  thoughts?

dhollinger:
@cdenneen I believe that was in reference to `rubocop`. While `rubocop` and `puppet-lint` checks are largely dependent on configuration and the current style guide standards (thus almost the same on all versions), Ruby and Puppet syntax checks should be considered version dependent as there are good chances that syntax will change between versions of Ruby and versions of Puppet.

cdenneen:
@dhollinger ok so maybe only part of @npwalker change for single rubocop?

npwalker:
Something to note, the travis matrix only runs the various checks on the newest ruby so I copied that here.  If there are valid reasons to run some of the checks other than spec on different versions of ruby then that should be corrected in travis and appveyor as well.  

However, I wouldn't think ruby versions shouldn't matter on their own but might in conjunction with a puppet version for something like puppet parser validate.

DavidS:
Since I've merged the other changes to gitlab-ci, this needs to be rebased and fixed up before I can push the button.

dhollinger:
@npwalker If a user is using any of the Puppet Ruby APIs like custom Facts, Types and Providers/Resource API, custom Ruby Functions, or simply extending Puppet with Ruby, then the Ruby version needs to have its syntax validated as there very well could be syntax changes between Ruby versions.

Correct me if I'm off/wrong on anything @DavidS

npwalker:
@dhollinger I took the latest updates to gitlab CI and they don't look right to me.  We're running things on ruby 2.4.1 and 2.4.4 but that doesn't seem like the right use of resources.  It also looks like we lost the syntax check that was there before.  

image

npwalker:
I think this PR should be closed since it no longer really applies after the changes that went into the template, however, I opened [PDK-981](https://tickets.puppetlabs.com/browse/PDK-981) to discuss further improvements to .gitlab-ci.yml.

PR #71, opened 2018-05-01
Update Gitlab template to be more configurable

dhollinger:
In using the template I found that the original `.gitlab-ci.yml.erb` file I contributed was too restrictive and decided to rework it to allow more customization.


dhollinger:
@cdenneen The new PR _can_ account for #44, I can update it now to include that if you'd like.

dhollinger:
@cdenneen Updated with simple matrix.

PR #72, opened 2018-05-01
Add quotes around the spec_helper's hiera_config

jarretlavallee:
Prior to this PR, the value for `spec_helper.rb` `hiera_config` was
not wrapped in quotes. It resulted in a failure when using the
`hiera_config` option as there is no `spec` object defined. This PR
adds quotes around the value in the template since it should always be a string.

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -34,7 +34,7 @@ end
 RSpec.configure do |c|
   c.default_facts = default_facts
   <%- if @configs['hiera_config'] -%>
-  c.hiera_config = <%= @configs['hiera_config'] %>
+  c.hiera_config = "<%= @configs['hiera_config'] %>"
   <%- end -%>
   <%- if @configs['strict_level'] -%>
   c.before :each do


npwalker:
@bmjen @DavidS thoughts?

npwalker:
@jarretlavallee what do you think?  

I'm inclined to say this is a bug that should just be fixed and anyone working around it will quickly update to no longer include quotes and we'll all be better off in the future not dealing with both cases.  But if need to keep compatibility then we could do some extra checking.

DavidS:
The main problem around changing this is that it forces organisations (and through FOSS modules, the entire world) to upgrade their templates/PDK together. Given that this is affecting the templates, and not day-to-day use, this is not a problem in this particular case (backaway).

jarretlavallee:
@DavidS Are there use cases for this value to be anything other than a string? It seems like the users that would be affected by this change are working around it by escaping quotes currently. If that is the case it seems like the behavior change may be beneficial in the long run. 

I agree with the issue with the single quotes as well, as the user could want some variable interpolation. I have updated it to double quotes for now and can change the logic to the results of this discussion.

Odiobill:
Ops sorry, didn't see that request and added (also) the same fix to mine as well.

PR #73, opened 2018-05-02
(PDK-967) Add an option to specify additional lines in Rakefile

jarretlavallee:
This adds an `extras` configuration option to allow for specifying arbitrary lines to the Rakefile. This is to enable adding additional requires and configuration options like
`MetadataJsonLint.options.strict_license`.

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -16,3 +16,7 @@ import '<%= item %>'
 <% checks.each do |check| -%>
 PuppetLint.configuration.send('<%= check %>')
 <% end -%>
+
+<%- [@configs['extras']].flatten.compact.each do |line| -%>
+<%= line %>
+<%- end -%>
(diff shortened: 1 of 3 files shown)


moduleroot/Rakefile.erb
```diff
@@ -16,3 +16,7 @@ import '<%= item %>'
 <% checks.each do |check| -%>
 PuppetLint.configuration.send('<%= check %>')
 <% end -%>
+
+<%- [@configs['extras']].flatten.compact.each do |line| -%>
```
DavidS: This will fail if there are no `extras` configured for a module. Either add a default empty array to the config_defaults, or check here for `nil`-ness.

jarretlavallee: @DavidS very good point. Sorry about that. The default array have been added.

jarretlavallee:
@DavidS Good point on the license. The license checks should be accomplished a different way, but it would still be worthwhile to add items to the RakeFile.

PR #67, opened 2018-05-01
Add a remove_includes configuration option

npwalker:
Prior to this commit, the user could not remove the default included checks
with configuration.

After this commit, the user can specify default checks that should be removed
in the remove_includes configuration setting in .sync.yml

This allows for a .sync.yml like the following that replaces check=spec with
check=parallel_spec.

```
.travis.yml:
  remove_includes:
    - env: CHECK=spec
    - env: PUPPET_GEM_VERSION="~> 4.0" CHECK=spec
      rvm:  2.1.9
  includes:
    - env: CHECK=parallel_spec
    - env: PUPPET_GEM_VERSION="= 4.6.1" CHECK=parallel_spec
      rvm: 2.1.9
    - env: PUPPET_GEM_VERSION="~> 4.0" CHECK=parallel_spec
      rvm: 2.1.9
```

PR #74, opened 2018-05-02
Add an option to remove default includes in .travis.yml and .gitignore

jarretlavallee:
This adds an `excludes` option to remove `includes` from the checks. This enables the .sync.yml to remove some of the default checks to do things like using parallel_spec instead of spec. 

This also adds the same option to the `.gitignore`.


npwalker:
I just got remove_includes added over in https://github.com/puppetlabs/pdk-templates/pull/67.  

I can imagine your solution might be a little more robust than mine.  However, I'm not sure if the exclude configuration name might not be super clear that it can remove default configuration.  

If this were to go in it should probably remove my remove_includes option on .travis.yml  if people find this to be a more robust solution.

jarretlavallee:
I did not see #67, so this PR is redundant as it provides essentially the same behavior. The only addition is the .gitignore. 

Closing this in favor of #67

PR #76, opened 2018-05-03
More Gitlab CI improvements

dhollinger:
Added following improvements:

* `.gitlab-ci.yml` documentation added to the README.md
* Added code to `.gitlab-ci.yml.erb` to allow for users to override the defaults completely if they so choose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,11 +47,24 @@ Gitlab CI uses a .gitlab-ci.yml file in the root of your repository tell Gitlab 
 
  Key            | Description   |
 | :------------- |:--------------|
+<<<<<<< HEAD
+<<<<<<< HEAD
 | override       |Defines whether your local `.sync.yml` will ignore the default values in pdk-templates. Defaults to `false`|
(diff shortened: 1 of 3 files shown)


dhollinger:
@cdenneen I'm thinking of how to do that still before I add anything. It would take some work as Gitlab CI doesn't support matrices the way Travis does. Gitlab CI operates in Pipelines (connected jobs) rather than a bunch of one-off "build" jobs (Just comparing to my experience with both).

cdenneen:
yeah so you'd still have something like
```
stages:
  - validate
  - metadata
  - unit
  - acceptance

ruby_versions:
  2.1.9:
  puppetversion: 4.10.11
  2.4.3:
  puppetversion: 5.5.1

```
and then some way of doing

for each stage
   for each ruby version
      for each puppet version

so the jobs come out as

validate puppet 4
validate puppet 5
metadata puppet 4
metadata puppet 5
unit puppet 4
unit puppet 5
acceptance puppet 4
acceptance puppet 5

PR #78, opened 2018-05-08
Delete .project template

DavidS:
Geppetto is long EOL, carrying this file is no longer useful to users, while causing many to try to get rid of it.

This fixes https://github.com/puppetlabs/pdk-module-template/issues/64


jearls:
isn't this the same as #58?

DavidS:
:man_facepalming:

PR #83, opened 2018-05-11
(PDK-981) Make .gitlab-ci look like .travis

npwalker:
Prior to this PR we created a pipeline that looked like: 

image

This PR creates a pipeline that looks like this: 

image

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -558,18 +558,12 @@ Gemfile:
     ruby_versions:
       '2.1.9':
         checks:
-          - 'lint validate metadata_lint'
           - spec
         puppet_version: '~> 4.0'
-      '2.4.1':
-        checks:
-          - 'lint validate metadata_lint'
-          - rubocop
-          - spec
-        puppet_version: '~> 5.0'
       '2.4.4':
         checks:
-          - 'lint validate metadata_lint'
+          - 'syntax lint metadata_lint'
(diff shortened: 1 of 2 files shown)


npwalker:
@dhollinger @bmjen @DavidS let me know what you think.

dhollinger:
@npwalker @DavidS Looks good to me

config_defaults.yml
```diff
@@ -558,18 +558,13 @@ Gemfile:
...
         puppet_version: '~> 4.0'
-      '2.4.1':
-        checks:
-          - 'lint validate metadata_lint'
-          - rubocop
-          - spec
-        puppet_version: '~> 5.0'
       '2.4.4':
         checks:
           - 'lint validate metadata_lint'
```
dhollinger: Just noticed this. Probably don't need validate and lint in the metadata_lint task if you're using `sytnax` and `lint` in another job

npwalker: Honestly I think rubocop, syntax, lint, validate, metadata_lint should all happen at the same time.  I'll don't see any reason to incur the cost of a new container to run them separately.  

What do you think?

dhollinger: I think it's good to keep `rubocop` separate for now since that's how I've seen it in most repos and internally. The `validate` task can be removed as I believe the `syntax` task covers that.

PR #88, opened 2018-05-14
(PDK-987) Save CI resources / reduce developer feedback time

npwalker:
Prior to this PR, we spent quite a bit of time spinning up containers and installing software instead of actually checking anything.  

After this PR, we get quicker feedback and save resources on our various free CI providers by running more checks at the same time instead of in new containers.  

Example of Travis before ( 311 seconds ): 
image

Example of Travis after ( 66 seconds ): 
image

Specifically Travis notes it only takes 2.89 seconds to run all checks:
image

Example Gitlab CI before: 
image

Example Gitlab CI After:
image


bmjen:
I like these changes. Before I merge though, I'd like @HelenCampbell or @pmcmaw to take a look and see if this will impact supported modules `.sync.yml` in a major way.

PR #90, opened 2018-05-16
Add travis testing to ensure template changes work

npwalker:
See an example of what it looks like here: 

https://travis-ci.com/npwalker/pdk-templates/builds/73507717

### DIFF
--- a/.travis/test_script.sh
+++ b/.travis/test_script.sh
@@ -1,0 +1,15 @@
+
+echo 'setup puppetlabs apt repo'
+wget https://apt.puppetlabs.com/puppet5-release-trusty.deb
+sudo dpkg -i puppet5-release-trusty.deb
+sudo apt-get update
+
+echo 'install pdk'
+sudo apt-get install pdk
+
+PR_DIR=$PWD
+
+pdk new module pr_new_modules --template-url=$PR_DIR --skip-interview
+pdk new module normal_module --skip-interview
+cd normal_module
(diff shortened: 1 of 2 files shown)


.travis/test_script.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo 'install pdk'
+sudo apt-get install pdk
+
+PR_DIR=$PWD
+
+pdk new module pr_new_modules --template-url=$PR_DIR --skip-interview
+pdk new module normal_module --skip-interview
+cd normal_module
+pdk convert --template-url=$PR_DIR --skip-interview --force
```
npwalker: Not sure what the expected behavior of `--skip-interview` on `pdk convert` is but it raised a question so I had to add `--force`.  

It makes me wonder if `--skip-interview` even does anything useful on `pdk convert`.

.travis/test_script.sh
```diff
@@ -0,0 +1,15 @@
+PR_DIR=$PWD
+
+pdk new module pr_new_modules --template-url=$PR_DIR --skip-interview
+
+pdk new module normal_module --skip-interview
+cd normal_module
+pdk convert --template-url=$PR_DIR --skip-interview --force
+cat convert_report.txt
```
npwalker: Not sure I need to output this convert_report.  The one below for difference from HEAD is more useful.

npwalker:
@bmjen did you also enable travis on this repo?

Issue #89, opened 2018-05-15
.sync.yml not working for .gitlab-ci.yml


Odiobill:
Looks like that the instruction provided in the README.md file for changing the content of .gitlab-ci.yml are not correct, or there is a bug. Currently I'm using this content for my .sync.yml:

```yaml
---
.gitlab-ci.yml:
  override: true
  custom:
    ruby_versions:
      '2.4.4':
        'puppet_version': '~> 5.5'

appveyor.yml:
  delete: true
```

The resulting .gitlab-ci.yml:

```yaml
---
stages:
  - test_2.4.1
  - test_2.1.9

before_script:
  - bundle -v
  - rm Gemfile.lock || true
  - gem update --system
  - gem update bundler
  - gem --version
  - bundle -v
  - bundle install --without system_tests

rubocop-2.4.1:
  stage: test_2.4.1
  image: ruby:2.4.1
  script:
    - bundle exec rake rubocop

syntax-2.4.1:
  stage: test_2.4.1
  image: ruby:2.4.1
  script:
    - bundle exec rake syntax lint

metadata-2.4.1:
  stage: test_2.4.1
  image: ruby:2.4.1
  script:
    - bundle exec rake metadata_lint

rspec-puppet-2.4.1:
  stage: test_2.4.1
  image: ruby:2.4.1
  variables:
    PUPPET_GEM_VERSION: ~> 4.0
    CHECK: spec
  script:
    - bundle update
    - bundle exec rake $CHECK

rubocop-2.1.9:
  stage: test_2.1.9
  image: ruby:2.1.9
  script:
    - bundle exec rake rubocop

syntax-2.1.9:
  stage: test_2.1.9
  image: ruby:2.1.9
  script:
    - bundle exec rake syntax lint

metadata-2.1.9:
  stage: test_2.1.9
  image: ruby:2.1.9
  script:
    - bundle exec rake metadata_lint

rspec-puppet-2.1.9:
  stage: test_2.1.9
  image: ruby:2.1.9
  variables:
    PUPPET_GEM_VERSION: ~> 4.0
    CHECK: spec
  script:
    - bundle update
    - bundle exec rake $CHECK
```

Moving "ruby_versions" outside "custom", changes something:

```yaml
---
.gitlab-ci.yml:
  override: true
  ruby_versions:
    '2.4.4':
      puppet_version: "~> 5.5.1"

appveyor.yml:
  delete: true
```

Result:

```yaml
---
stages:
[...]

PR #91, opened 2018-05-16
Add support for custom docker images for .gitlab-ci.yml

Odiobill:
Following the discussion we had on Slack and https://github.com/puppetlabs/pdk-templates/issues/89 I decided to take a closer look to the source code. Adding support for the custom docker image that I need turns out to be pretty straightforward.

I also added quotes around the hiera config file path that you can specify for spec/spec_helper.rb (otherwise the slashes can be confused for a regular expression, I guess).

### DIFF
--- a/moduleroot/.gitlab-ci.yml.erb
+++ b/moduleroot/.gitlab-ci.yml.erb
@@ -51,7 +51,7 @@ before_script:
   stage: syntax
 <%       end -%>
 <%       if options['docker_image'] -%>
-  image: <%= docker_image %>
+  image: <%= options['docker_image'] %>
 <%       else -%>
   image: ruby:<%= ruby_version %>
 <%       end -%>


moduleroot/.gitlab-ci.yml.erb
```diff
@@ -50,7 +50,11 @@ before_script:
 <%       else -%>
   stage: syntax
 <%       end -%>
+<%       if options['docker_image'] -%>
+  image: <%= options['docker_image'] %>
+<%       else -%>
```
npwalker: This seems a little strange since it's happening instead the loop for adding ruby versions but you're overriding the docker_image which is the thing that provides the ruby version.  

Are you using docker images that also provide the needed ruby versions otherwise it seems like this would be confusing?

npwalker:
@Odiobill I think an improvement would be to just have a different setting that allows the user to specify a matrix in their sync.yml and have it override the default of looping through the ruby versions.  However, that doesn't preclude this option.  

You'll probably need to rebase and squash this PR since it's conflicting now and a different PR was merged for the quotes on hiera.yml that uses double quotes.

Odiobill:
Yes, you are right. Unfortunately I didn't notice the other request before submitting mine. I just submitted another PR for that single change.

PR #93, opened 2018-05-17
Make gitlab cache configurable

dhollinger:
Extended the cache settings that @npwalker added to make the setting configurable and move the default settings from the template into the `config_defaults.yml` file


npwalker:
If I'm understanding correctly this allows a user to add additional paths to be cached via sync.yml.  

Looks good to me.

dhollinger:
@bmjen @DavidS - review and/or merge when you get a chance?

PR #95, opened 2018-05-17
(MODULES-7184) replace parallel_spec with spec

eputnam:
parallel_spec and bundler 1.16.2 are not playing nicely. The new bundler version was released yesterday and now checks that use parallel_spec are failing. Not sure if this is a permanent solution or not but it is definitely the easiest solution at this point and it should get things green again. Another equally easy option would be to remove the bundler update from the before_install array, but that seems like it would have much more of an impact.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -49,8 +49,8 @@
     script: bundle exec rake beaker
   includes:
     - env: CHECK="syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop"
-    - env: CHECK=parallel_spec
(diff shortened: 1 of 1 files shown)


npwalker:
@eputnam what seems to be the issue?  I run parallel_spec with 1.16.2 without issue.

eputnam:
There may be a third factor: rubygems 2.7.6 

And then of course any number of things that could be going on with the box from travis.

Some examples:
https://travis-ci.org/puppetlabs/puppetlabs-zfs_core/jobs/379959823
https://travis-ci.org/puppetlabs/puppetlabs-apache/jobs/380280592

npwalker:
I put up https://github.com/puppetlabs/pdk-templates/pull/96.

PR #98, opened 2018-05-18
Add support for custom docker images

Odiobill:
New pull request for this specific change after the comments on the previous one.

### DIFF
--- a/moduleroot/.gitlab-ci.yml.erb
+++ b/moduleroot/.gitlab-ci.yml.erb
@@ -50,11 +50,6 @@ before_script:
 <%       else -%>
   stage: syntax
 <%       end -%>
-<%       if options['docker_image'] -%>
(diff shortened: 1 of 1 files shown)


bmjen:
CC @npwalker @jarretlavallee

npwalker:
@bmjen I think this partially solves the problem and allows adding in your own docker image but since the loop is based on ruby version and the ruby version was coming from the docker image it seems like an unreliable solution.  

Granted a more reliable solution would probably result in a breaking change or we'd have to have two separate code paths based on whether ruby_version was supplied ( or something like that).

ghoneycutt:
This PR seems very straight forward. Why not merge it or close it.

scotje:
Seems like this was resolved by #137, please let us know if not.

PR #105, opened 2018-06-02
Several beaker-related fixes from #104

### LINKED ISSUE #104: gitlab-ci beaker
@dhollinger  @npwalker 

Few things:

- [x] there is no rake task of `acceptance` anymore... needs to be `beaker`
- [x] the `bundle install` for acceptance job needs to be `bundle install --with system_tests` otherwise the bundle install does nothing but uses whats in cache and beaker never gets installed.
- [x] image should allow for registry in `.sync` or `config_defaults` and if set prepend with the registry_host/ 
- [x] beaker variables (and probably any other variables needs to convert boolean to string)
```yaml
---
.gitlab-ci.yml:
  custom:
    beaker:
      variables:
        BEAKER_debug: 'true'
```
converts to `BEAKER_debug: true` and then fails ci lint as it needs to be `BEAKER_debug: 'true'`
- [ ] beaker acceptance job missing default image and/or gitlab-ci.yml doesn't have a default set so falls back to runners configured default image which in the case of gitlab.com is ruby:2.5 but in my local hosted gitlab there is no default so it fails:
```
Running with gitlab-runner 10.8.0 (079aad9e)
  on gitlab-ci-runner-1 e1b7b7f8
ERROR: Preparation failed: no image specified and no default set in config
Will be retried in 3s ...
ERROR: Preparation failed: no image specified and no default set in config
Will be retried in 3s ...
ERROR: Preparation failed: no image specified and no default set in config
Will be retried in 3s ...
ERROR: Job failed (system failure): no image specified and no default set in config
```
- [ ] outside the scope of this in general but `gem install nokogiri` even cached will recompile during acceptance job I believe because that needs to be installed with `--prune` flag or something if I recall reading that somewhere.
- [ ] currently by having 2 spec and lint jobs running in same stage there is absolutely NO benefit to caching in gitlab-ci. In order for caching to make sense is the use of multiple stages. (technically right now the last syntax stage job is the one winning in the "cache.zip" upload... since they are overwriting each other.) I thought we originally discussed needing a "quick" job as the first stage like linting which could create the cache and speed up the spec/acceptance jobs downstream? Also if your going to fail linting, validate, maybe even rubocop it's probably not worth installing all fixtures and full spec testing suite since some of those can be massive and fail early in previous stage? Maybe even consider a "prep" job that just does the `bundle install --with development` and creates the cache (seems like waste for container so still recommend moving the linting to first stage)? It's possibly the lint was supposed to be first then at that point the ERB is just a typo and the spec ones should be `unit` stages
- [ ] setting override removes `syntax` and `unit` from the stages but not the jobs. I believe if there are jobs calling a stage that is no longer defined this will break ci linting. So maybe if override the jobs get removed as well as the stages?

dhollinger:
Addresses #104

### DIFF
--- a/moduleroot/.gitlab-ci.yml.erb
+++ b/moduleroot/.gitlab-ci.yml.erb
@@ -42,6 +42,11 @@ before_script:
 
 <% if configs['ruby_versions'] -%>
 <%   configs['ruby_versions'].each do |ruby_version, options| -%>
+<%     if options['image'] -%>
+<%       image = options['image'] -%>
+<%     else -%>
+<%       image = 'ruby' -%>
+<%     end -%>
 <%     options['checks'].each do |check| -%>
 <%= check %>-Ruby <%= ruby_version %>-Puppet <%= options['puppet_version'] %>:
 <%       if check == 'spec' -%>
@@ -49,7 +54,7 @@ before_script:
 <%       else -%>
   stage: syntax
 <%       end -%>
-  image: ruby:<%= ruby_version %>
+  image: <%= image %>:<%= ruby_version %>
   script:
     - bundle exec rake <%= check %>
   variables:
--- a/moduleroot/.gitlab-ci.yml.erb
+++ b/moduleroot/.gitlab-ci.yml.erb
@@ -14,7 +14,7 @@ stages:
   - unit
 <% end -%>
 <% if configs['beaker'] -%>
-  - acceptance
+  - beaker
 <% end -%>
 <% if configs['custom_stages'] -%>
 <%   configs['custom_stages'].each do |stage| -%>
@@ -24,7 +24,7 @@ stages:
 <% if configs.has_key?('global_variables') -%>
 variables:
 <%   configs['global_variables'].each do |key, value| -%>
-  <%= key %>: '<%= value %>''
+  <%= key %>: '<%= value %>'
 <%   end -%>
 <% end -%>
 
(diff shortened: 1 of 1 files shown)


dhollinger:
This superscedes #98 if @cdenneen and @ Odiobill approve

cdenneen:
I've marked 4 of 8 completed in #104 
Think there is 1 that really can't be fixed but other 2-3 can probably still be fixed here if want to have everything in one. (While we wait for @odiobill to approve the custom `image` changes)

dhollinger:
@cdenneen I'm not sure what the best way to go about setting up the "default" job would look like for beaker. It seems like beaker configs are moving away from that and more to `beaker-host-generator` instead (at least for Docker jobs). But even then, each user's `default` job will be potentially unique to them.

Anything we do within this repo should conform to the best practices put forth/expected by Puppetlabs with the ability to be modified as needed by the user. The `gitlab-ci.yml`, `travis.yml`, `appveyor.yml`, etc should all be creating the same jobs as close as possible by default.

Odiobill:
@cdenneen is there anything else I should do to proceed further?

cdenneen:
@Odiobill do you agree with @dhollinger change for your custom image request? #98

cdenneen:
CC @DavidS

bmjen:
CC @npwalker @jarretlavallee 

These changes LGTM, but I was hoping you'd take a look to see if any of these gitlab changes would impact you.

PR #109, opened 2018-06-09
fix stage bug

davidvanlaatum:
unit tests were ending up in the syntax stage


cdenneen:
I've asked @dhollinger to add this to the #105 PR just to keep everything together. 
I'm curious if there is a way we can make this spec/parallel_spec a variable/boolean (`parallel: true`, `coverage: true`), or something to enable/disable parallel and use something like `check =~ /spec/` to account for both... if using:

```
RSpec.configure do |c|
  c.after(:suite) do
    RSpec::Puppet::Coverage.report!(90)
  end
end
```
parallel causes issues with this.

Does no one else use `Coverage.report`?

bmjen:
CC @npwalker @jarretlavallee

PR #111, opened 2018-06-18
(feature) add github changelog generator template

tphoney:


tphoney:
@DavidS ? what do you think ?

moduleroot/Rakefile.erb
```diff
@@ -12,11 +13,43 @@ import '<%= item %>'
...
+
 <% checks = @configs['default_disabled_lint_checks'] + ( @configs['extra_disabled_lint_checks'] || [] ) -%>
 <% checks.each do |check| -%>
 PuppetLint.configuration.send('disable_<%= check %>')
 <% end -%>
 
+if Bundler.rubygems.find_name('github_changelog_generator').any?
+  GitHubChangelogGenerator::RakeTask.new :changelog do |config|
+    config.user = "#{field_from_metadata('author')}"
+    config.project = "#{field_from_metadata('name')}"
```
DavidS: These two GCG settings should match up with github, which is not necessarily the case with `metadata`. See e.g. [puppetlabs-ntp/metadata.json#L4](https://github.com/puppetlabs/puppetlabs-ntp/blob/4fd50940f109023fbb3347571c98f60bd6a59e53/metadata.json#L4)

Allow overriding through `.sync.yml`?

config_defaults.yml
```diff
@@ -503,6 +503,11 @@ Gemfile:
         condition: "Gem::Version.new(RUBY_VERSION.dup) < Gem::Version.new('2.1.0')"
       - gem: fast_gettext
         condition: "Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.1.0')"
+      # add the changelog gem to puppet-module-gems, when there is a release
+      - gem: 'github_changelog_generator'
+        git: 'https://github.com/skywinder/github-changelog-generator'
+        ref: 'master'
```
DavidS: * This introduces a dependency that's not part of the PDK package, which will break offline usage
* Instead of a branch name, this should list a SHA to avoid any update to GCG break our users.

moduleroot/Rakefile.erb
DavidS:
```
david@davids:~/git/david-testmod$ pdk bundle exec rake changelog
pdk (INFO): Using Ruby 2.4.4
pdk (INFO): Using Puppet 5.5.1
Fetching tags...
rake aborted!
Octokit::NotFound: GET https://api.github.com/repos/david/david-testmod/tags?per_page=100: 404 - Not Found // See: https://developer.github.com/v3/repos/#list-tags
```

Exactly what happened to me. See https://github.com/DavidS/david-testmod/blob/7c9df6bf827168b448d04d51fceb89783d701544/metadata.json#L4 which was prefilled from my username, but doesn't match my github handle.

tphoney: groan, should be fixed in ntp. and overrides FTW

tphoney:
@DavidS ready for review again

tphoney:
@DavidS 👍  love it, just need a merge now

PR #112, opened 2018-06-19
Add system_tests and add beaker-aws environment

cdenneen:
Now that https://tickets.puppetlabs.com/browse/BKR-1464 has been resolved should be able to enable `system_tests` group.
Also due to issue with https://tickets.puppetlabs.com/browse/BKR-1483 it would be beneficial to allow `beaker-aws` to have environment variable for override

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -529,25 +529,27 @@ Gemfile:
           - mswin
           - mingw
           - x64_mingw
-    # ':system_tests':
-    #   - gem: 'puppet-module-posix-system-r#{minor_version}'
-    #     platforms: ruby
-    #   - gem: 'puppet-module-win-system-r#{minor_version}'
-    #     platforms:
-    #       - mswin
-    #       - mingw
-    #       - x64_mingw
-    #   - gem: beaker
-    #     version: '~> 3.13'
-    #     from_env: BEAKER_VERSION
-    #   - gem: beaker-abs
-    #     from_env: BEAKER_ABS_VERSION
-    #     version: '~> 0.1'
-    #   - gem: beaker-pe
-    #   - gem: beaker-hostgenerator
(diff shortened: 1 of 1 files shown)


cdenneen:
@bmjen have you had time to review this? I don't think enabling the Gem group will force anyone to enable `beaker` job. Gem dependencies can get intertwined and for someone to make sure of all the work to add jobs for beaker to CI that's been added to this, want to make sure PDK controls those gem dependencies.

bmjen:
@cdenneen I haven't yet. Will likely discuss it with the rest of the team at our bi-weekly grooming tomorrow. My biggest concern is that... adding this `system_tests` group will now force all new modules to add these dependencies, regardless of their need or want for acceptance testing and use of beaker. We'd be in a situation where users who don't want beaker will be in the precarious position to have to remove the deps in a `.sync.yml`. Additionally... there will be packaging concerns. Adding these gems to the template will result in our packages containing all of the beaker gem deps in the installer, if we added a special condition to not package them, then PDK users that install into an airgapped environment won't be able to run because they won't be able to download the gems online.

One alternative that could work is wrapping this entire section in a conditional called `include_system_tests` and default it to false. Then you and whoever else that wants to include the `system_tests` would add `include_system_tests: true` to your `.sync.yml`.

rtib:
Instead of a new switch, `system_tests` group may be included automatically if a non-empty `docker_sets` is present.

rtib:
@bmjen while I understand your arguments, my opinion is, that including `system_tests` by default (probably along with merging #172) will lower the threshold to enable beaker, which will lead to more mature puppet modules.

scotje:
Thank you for your contribution, we are planning to move another direction in regards to enabling acceptance testings via PDK so we are going to close this PR.

PR #115, opened 2018-06-21
add ability to support BEAKER_TESTMODE

b4ldr:
This allows one to set the BEAKER_TESTMODE variable using the .sync file.  if no value is set then a default of `apply` is used


b4ldr:
@DavidS i have added some text however none of the other hash keys are described so i wasn't sure of the best way to shoe horn theses in.  happy to reformat if you have some ideas

b4ldr:
@DavidS sorry i reset your review as i accidentally pushed to this branch.

PR #118, opened 2018-06-25
(PDK-1044) changelog functions if changelog task

tphoney:
Only run the changelog helpeer functions if we are running the changelog task


DavidS:
This will not fix the hard failure when `CHANGELOG_GITHUB_TOKEN` is unset

tphoney:
attempt no.2 @DavidS

PR #120, opened 2018-07-06
(PDK-1066) Use ruby 2.4.4 with travis

vinzent:
This aligns with the ruby version provided by the pdk.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -35,14 +35,14 @@
   required: *ignorepaths
 .travis.yml:
   ruby_versions:
-    - 2.4.1
+    - 2.4.4
   global_env:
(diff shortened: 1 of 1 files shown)


vinzent:
```
convention: rubocop: lib/puppet/provider/testprovider/testprovider.rb:4:60: Layout/SpaceBeforeFirstArg: Put one space between the method name and the first argument.
```
is not related to this change.

this class line triggers the warning:

```ruby
require 'puppet/resource_api/simple_provider'

# Implementation for the testprovider type using the Resource API.
class Puppet::Provider::Testprovider::Testprovider < Puppet::ResourceApi::SimpleProvider
  def get(_context)
    [
...
```

doesn't make sense to me.

rodjek:
That rubocop failure was due to a bug in one of rubocop's dependencies.

PR #124, opened 2018-07-22
remove concat_basedir from default-facts

### LINKED ISSUE #122: Default facts uses concat_basedir
Default facts uses concat_basedir which is a deprecated fact as part of the concat module.  Also it uses a linux only path.

Why is this included in the default facts for all modules created by the PDK?

TheMeier:
fixes #122


TheMeier:
As @glennsarti pointed out in #122 concat_basedir was removed a long time ago. I tested it with puppetlab-apache module and its tests run just fine without this fact.

PR #129, opened 2018-08-08
(CLOUD-2063) Moving variables for use in Gem versions

HelenCampbell:


HelenCampbell:
For using the variables in the Gems (for versions) we need them defined at the top of the file - although there's further questions around this, I think perhaps the version cannot be set as a variable as it seems to be setting it as a String.

PR #136, opened 2018-09-04
Implement notifications configuration for .travis.yml

### LINKED ISSUE #119: Can't customize .travis.yml notifications
While looking at converting a module via `pdk convert` I noticed that there was no way to define my on notification rules via `.sync.yml`. The PDK forces email to false with no way to override that and also does not provide any way to add other notification methods. In my case, I need to add in a section for Slack.

nmaludy:
Closes #119 

This PR implements an override in `.travis.yml` for the `notifications` section so repos can send notifications to things like Slack.


ardrigh:
I have need for similar functions. It would be nice if these PRs could be reviewed.

moduleroot/.travis.yml.erb
```diff
@@ -88,7 +88,18 @@ branches:
 <%   end -%>
 <% end -%>
 notifications:
-  email: false
+<% notifications_arr = ((@configs['notifications'].to_a || []) - (@configs['remove_notifications'].to_a || [])) -%>
```
scotje: Does this need to guard against `@configs['notifications']` possibly being `nil`? I realize it has a default, but someone maybe could override it to `nil`?

moduleroot/.travis.yml.erb
```diff
@@ -88,7 +88,18 @@ branches:
...
+<% if notifications_arr.any? -%>
```
scotje: I see elsewhere in the file we currently use `.any?` without a block as a conditional, but I think if we were applying rubocop rules here it would be:

```suggestion
<% unless notifications_arr.empty? -%>
```

moduleroot/.travis.yml.erb
scotje: Was there a situation where this value was not an array but could be safely coerced into an array? If not, we can just drop to:

```suggestion
<% notifications_arr = ((@configs['notifications'] || []) - (@configs['remove_notifications'] || [])) -%>
```

nmaludy: @scotje Notifications is a hash, but in order to provide a "difference" between the `@config['notifications']` and `@config['remove_notifications']` i converted to an array to make this difference logic easier.

moduleroot/.travis.yml.erb
nmaludy: Fixed.

ardrigh:
I think this code creates a bug in `.travis.yml` if you need to set `email: false`.

In `.sync.yml` this adds the slack channel notification and removes `email:false` in `.travis.yml`
```yaml
  notifications:
    - slack:
        secure: BLAHBLAHBLAHBLAHBLAHBLAH
```

I tried a few variations of `notifications` but none of them add `email:false` as expected.

Any ideas?

scotje:
@ardrigh I was able to configure both `email: false` and a `slack` notification with the following in my `.sync.yml`:

```yaml
notifications:
  email: false
  slack:
    secure: foozle
````

and running `pdk update`. Does that config work for you?

ardrigh:
@scotje sorry I haven't tested this for a while. 

I just tried PDK 1.9.1 and I have the format working now, so it's not trying to remove the secure notification for Slack any longer.

PR #141, opened 2018-09-19
(MODULES-7742) - Update decoration rubocop rule to ignore spec folder

david22swan:


david22swan:
Throws a warning on non i18n modules but still works.

PR #143, opened 2018-09-20
(FM-7397) - Enable puppet6 testing in travis

pmcmaw:


pmcmaw:
screen shot 2018-09-20 at 11 38 23 am
screen shot 2018-09-20 at 11 38 14 am

I can confirm these changes have passed on motd and package. There has been failures on stdlib however these are expected. The final configuration is exactly what we are looking for.

Issue #154, opened 2018-10-25
Appveyor template builds branches + PR's producing a double build


TraGicCode:
Currently, i have my puppet module setup to build on both appveyor & travis-ci.  When i create a PR .2 builds for appveyor kick off.  1 for the branch and 1 for the merged PR.  Travis doesn't do this because it has already been configured to handle this and only perform one build.  Below is the PR that shows this double build happenings

https://github.com/TraGicCode/tragiccode-azure_key_vault/pull/25

![image](https://user-images.githubusercontent.com/6224270/47496517-9842d200-d81c-11e8-85e8-560fbab84159.png)

**NOTE: This double build only happens when you own the public github repo and cannot create a fork**

TraGicCode:
This was closed from #155

PR #155, opened 2018-10-25
(GH-154) Fix appveyor double build

TraGicCode:
The problem is described in #154.  This commit ensures only master and pr branches get build.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -88,6 +88,8 @@ appveyor.yml:
       CHECK: parallel_spec
   test_script:
     - bundle exec rake %CHECK%
+  branches:
+    - master
 Rakefile:
   default_disabled_lint_checks:
     - 'relative'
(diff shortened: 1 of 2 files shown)


DavidS:
Please copy the logic from https://github.com/puppetlabs/pdk-templates/blob/bcb7883e007456b3151f0136cc12b5d853927c25/moduleroot/.travis.yml.erb#L77-L89 and the defaults from https://github.com/puppetlabs/pdk-templates/blob/bcb7883e007456b3151f0136cc12b5d853927c25/config_defaults.yml#L61-L63 to keep the two in sync.

TraGicCode:
Hey @DavidS ,

I will work on making this consistent with how travis.

Thanks for the feedback!

TraGicCode:
Here is a pr showing the double build is gone when i point the puppet module at the pdk-template to my fix.

https://github.com/TraGicCode/tragiccode-azure_key_vault/pull/37

PR #156, opened 2018-10-25
(PDK-1210) setting inherited const_defined lookup to false

Thomas-Franklin:
If there is a top level module with same name as a type which gets loaded before the type gets defined.

This will trigger a false positive on the `last_module.const_defined?` check causing an `uninitialized constant` error.

This change will no longer do a recursive check and only care about the local namespace this way avoiding the false positive.

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -44,7 +44,7 @@ end
 def ensure_module_defined(module_name)
   module_name.split('::').reduce(Object) do |last_module, next_module|
     last_module.const_set(next_module, Module.new) unless last_module.const_defined?(next_module, false)
-    last_module.const_get(next_module)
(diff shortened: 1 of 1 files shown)


DavidS:
See also https://github.com/puppetlabs/puppet-resource_api/pull/132 and https://github.com/puppetlabs/puppet-resource_api/pull/134.

davidmalloncares:
@Thomas-Franklin - has there been any movement on this one recently?

Thomas-Franklin:
> @Thomas-Franklin - has there been any movement on this one recently?

@davidmalloncares it was approved by @scotje, but not sure if it needs more than one approval. I'll update the branch with master and have it ready anyway.

scotje:
Seeing if @bmjen or @rodjek can take a look, we typically do 2 reviews before merge.

PR #169, opened 2018-12-04
(PDK-1202) Generate a .puppet-lint.rc based on Rakefile

rodjek:
Our current documented way of disabling puppet-lint checks is to use
`.sync.yml` to specify a list of checks which is then used when
generating the `lint` task in the `Rakefile`. Unfortunately, while this
does take effect when running the tests in CI, it does not similarly
change the behaviour of `pdk validate` which calls `puppet-lint`
directly instead of using the rake task.

This PR adds a new template for `.puppet-lint.rc` which gets built from
the values specified for the `Rakefile` template. I have deliberately
not added any new config keys specific to this template to reduce the
chance of the puppet-lint configuration from getting out of sync again.

This PR can not be merged until puppetlabs/pdk#605 is merged.


rodjek:
The test is failing because we have Travis configured to use the latest release package to run the tests (which doesn't include the change). I'll update the test to use the latest nightly build instead.

PR #173, opened 2019-01-03
(FM-7659) - Fix to pin Bundler for Puppet 4 Testing

david22swan:
New version of Bundler does not support ruby versions older than 2.3 or RubyGem versions older than 3.0.0

### DIFF
--- a/moduleroot/.travis.yml.erb
+++ b/moduleroot/.travis.yml.erb
@@ -17,6 +17,9 @@ addons:
 <%   end -%>
 <% end -%>
 before_install:
+  - if [ $BUNDLER_VERSION ]; then
+      gem install -v $BUNDLER_VERSION bundler --no-rdoc --no-ri;
+    fi
   - bundle -v
   - rm -f Gemfile.lock
   - gem update --system $RUBYGEMS_VERSION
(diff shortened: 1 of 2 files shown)


pmcmaw:
Include ticket number.

pmcmaw:
@Thomas-Franklin if you are happy enough with these changes, would you mind merging.

moduleroot/.travis.yml.erb
```diff
@@ -17,6 +17,9 @@ addons:
 <%   end -%>
 <% end -%>
 before_install:
+  - if [ $BUNDLER_VERSION ]; then
+      gem install -v $BUNDLER_VERSION bundler --no-rdoc --no-ri;
```
joshcooper: We recently discovered  bundler 2 no longer accepts `--no-rdoc` and `--no-ri` in favor of `--no-document`. Apparently the latter two were deprecated a long time ago, but only in the help text.

pmcmaw: Thanks for the heads up @joshcooper :-)
When we drop Puppet 4 testing, we will be removing the bundler version pin and this code block therefore we should avoid running into any issues.

PR #175, opened 2019-01-03
(MAINT) Update JSON pinning to cover Ruby 2.4.5 and future Rubies

scotje:
This updates the generated Gemfile to avoid it trying to resolve JSON to 2.1.0 on Ruby 2.4.5 (and future 2.4.x releases of Ruby). It's possible we might have to tweak the ranges if Ruby embeds a different JSON version in a future release but that seems more rare than us just bumping the Ruby patch version.


scotje:
Good call, I'll update

PR #178, opened 2019-01-09
(MODULES-8444) - Removal of Puppet 4 testing from travis and appveyor

david22swan:


david22swan:
@glennsarti Could you take a look at this please. We are currently dropping all puppet 4 testing and since this will effect the cross platform modules and any windows modules that utilize the pdk.

jpogran:
@puppetlabs/windows for visibility

glennsarti:
@HelenCampbell @david22swan I'm mostly in favour of removing Puppet 4 testing, but I would expect there to be some documentation (or link) somewhere to assist people to re-add Puppet 4 testing if they need it because;

*  This repo affects almost all PDK users (IMO most people won't use a custom template location), not just Puppet (the company) people
* This repo isn't versioned or tagged so there's no concept of breaking changes.  In this case documentation seems like a good way to capture this.

HelenCampbell:
Merging due to updated documentation on the back of Glenns comment and windows approval of the change.

natemccurdy:
To anyone finding this PR because their Puppet 4 builds are failing in TravisCI and an update PDK isn't helping:

I got mine working by adding the following to my .sync.yml then running `pdk update`. The trick is setting RUBYGEMS_VERSION to a version old enough that it supports Ruby 2.1.9. This works as of PDK version 1.10.0.

```yaml
.travis.yml:
  includes:
    - env: PUPPET_GEM_VERSION="~> 4.0" CHECK=parallel_spec RUBYGEMS_VERSION=2.7.8
      rvm: 2.1.9
      stage: spec
  remove_includes:
    - env: PUPPET_GEM_VERSION="~> 6.0" CHECK=parallel_spec
      rvm: 2.5.3
      stage: spec
```

In my case, my module supports Puppet 4 and 5, but not 6. I want to use the latest version of the PDK, but don't need to add tests for versions of Puppet my module doesn't explicitly support.

PR #184, opened 2019-02-12
(PDK-1269) Add plan template

npwalker:


object_templates/plan.erb
```diff
@@ -0,0 +1,16 @@
...
+# You can write Puppet Plans in puppet code
+#
+# Learn more at: https://puppet.com/docs/bolt/1.x/writing_plans.html
+
+plan <%= name %> (
+  String[1]   $parameter1,
+  TargetSpec  $parameter2,
+) {
+
+  run_task('modulename::task_name', $parameter1, some_parameter_name => $parameter2)
```
bmjen: I'll have to check if we pass in the module name to the config hash, would be nice to fill in the `modulename::task_name` part.

npwalker: This is just some made up boiler plate of how someone would put a task in a plan.  However, up on line 10 it only passes though the name of the plan but without namespacing and we really have to fix that.  

https://github.com/puppetlabs/pdk-templates/pull/184/files#diff-41c11f3567ed263cffeae409027787f8R10

I think what the template for a plan should look like is something I'd want the bolt maintainers to provide.  Since it can run tasks, scripts, commands, puppet apply there's no one size fits all template but I think we want to get a small snippet of each ideally so the user can copy /paste.

bmjen: Yep, I agree. I'll merge this to get something boilerplate in, and we can address more codified templates of plans, tasks, and tests with the Bolt team.

Issue #198, opened 2019-03-04
Fix documentation of a resource API type to match documentation


ghoneycutt:
https://puppet.com/docs/puppet/6.3/puppet_strings_style.html#example-resource-api-type

PR #199, opened 2019-03-04
(GH-198) Fix documentation of resource API type

ghoneycutt:


ghoneycutt:
Fixes #198

object_templates/provider_type.erb
```diff
@@ -3,8 +3,16 @@ require 'puppet/resource_api'
 Puppet::ResourceApi.register_type(
   name: '<%= name %>',
   docs: <<-EOS,
-      This type provides Puppet with the capabilities to manage ...
-    EOS
+@summary Fancy new type
```
da-ar: `@summary a <%= name %> type`

object_templates/provider_type.erb
```diff
@@ -3,8 +3,16 @@ require 'puppet/resource_api'
...
+@example Fancy new example
```
da-ar:
```
@example 
<%= name %> { 'foo':
ensure: present,
name: 'foo'
}
```

object_templates/provider_type.erb
```diff
@@ -3,8 +3,16 @@ require 'puppet/resource_api'
...
+apt_key { '6F6B15509CF8E59E6E469F327F438280EF8D349F':
+ source => 'http://apt.puppetlabs.com/pubkey.gpg'
+}
+
+This type provides Puppet with the capabilities to manage ...
+**Autorequires**:
```
da-ar: I would be inclined to remove the "autorequires" section

ghoneycutt: I've found the autorequires to be very helpful. We use it extensively for the sensu module.

https://github.com/sensu/sensu-puppet/blob/sensu2/REFERENCE.md

object_templates/provider_type.erb
ghoneycutt: Good idea.. I just copy/pasted from the official docs

object_templates/provider_type.erb
ghoneycutt: We could foo it up or use a real example like apt_key. I'm not particularly attached to either.

DavidS: As a help to the developer, this should be as close as possible to the final result as we can get it, so `@example <%= name %> { 'foo': ...` is much better than something with `apt_key`, as it will be technically correct at the time of rendering, and not confuse people, e.g. in the generated documentation.

object_templates/provider_type.erb
DavidS: In the same vein as above, the template should not document support for something it doesn't actually implement. As a compromise we can add a "Further reading for developers" section here, that points out important things that folks should consider next after `pdk new provider`.

ghoneycutt:
@scotje I completed @da-ar 's request though it still shows as a change request in red. This is good to merge.

Issue #205, opened 2019-03-05
Hiera should not default to using common


ghoneycutt:
common makes sense for a control repo, not for module data. The 'common' value is the default value specified by the class parameter.

binford2k:
Counterpoint: that means that the parameter will default to a value coming from two different sources of truth. I'd have to look at Hiera and validate that no value is returned, and then look at the class declaration to see if there was a default specified. By setting the 'common' value in `common.yaml` and declaring no default in the class signature, I've only got to look at one source of truth.

ghoneycutt:
Ben, I'd like to agree but the tooling is not there and neither is the community. The pattern is not widely used in the community. If you do not specify a default value, puppet-strings will remove the default value from the REFERENCE.md which makes the module harder to understand. If you look at the code it would seem that the parameter must be specified because it has no value.

ghoneycutt:
If you take a look at PR #206 this is more inline with what most modules need, which is the ability to specify different data per osfamily and major release of each osfamily.

ghoneycutt:
@binford2k you there?

ghoneycutt:
@binford2k could you provide an example where a module uses `common.yaml` for data instead of default values for class parameters? Perhaps I'm misunderstanding you.

alexjfisher:
eg https://github.com/puppetlabs/puppetlabs-ntp/blob/master/data/common.yaml ?

(I'm not necessarily saying I like this pattern.  I also find the puppet-strings limitation a big negative)

rtib:
I don't like having parameter defaults in hiera either, especially within modules. As the PR is touching the initial hiera.yaml only, it won't affect any modules being updated, but forces module developers to think before dumping lots of default values blindfold into common.yaml, only because it's there.

ghoneycutt:
@rodjek wrt the comments above, could you please merge my PR #206. If Puppet feels strongly regarding `common.yaml` they just can add it back at the end. Having platform support in Hiera is super basic for any module.

rodjek:
@ghoneycutt to keep everyone happy can you modify your PR to reintroduce `common.yaml` at the lowest level of the hierarchy?

LongLiveCHIEF:
[...]

PR #206, opened 2019-03-05
(GH-205) Hiera to use better defaults for modules

ghoneycutt:


ghoneycutt:
Fixes #205

scotje:
@rodjek and/or @glennsarti you want to give this a once-over? See my approval comment, there is also a lot more discussion and context on #205.

glennsarti:
+1 Given this is in `_init`, even if it's slightly wrong it's a small "blast radius"

PR #209, opened 2019-03-05
Remove gems that are no longer needed for Puppet 5 or Puppet 6

ghoneycutt:


bmjen:
@ghoneycutt Thanks for the contribution here, and it will eventually get cleaned up. But for now I'm going to close this PR because we still need to maintain these gem/version dependencies until we remove 2.1.9 from PDK.

PR #213, opened 2019-03-08
Remove leading whitespace from README template

montaguethomas:
When generated, the README.md will have a leading newline.

### DIFF
--- a/moduleroot_init/README.md.erb
+++ b/moduleroot_init/README.md.erb
@@ -1,4 +1,4 @@
-<% name = @configs['module_metadata']['name'].split('-').last rescue '[modulename]' %>
+<%- name = @configs['module_metadata']['name'].split('-').last rescue '[modulename]' -%>
 # <%= name %>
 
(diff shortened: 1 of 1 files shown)


moduleroot_init/README.md.erb
```diff
@@ -1,4 +1,4 @@
-<% name = @configs['module_metadata']['name'].split('-').last rescue '[modulename]' %>
+<%- name = @configs['module_metadata']['name'].split('-').last rescue '[modulename]' -%>
```
ghoneycutt: I don't believe you want `<%-`Just `-%>` at the end to remove newlines.

Don't see why we want this line at all though since we have `# name` below it.

montaguethomas: The reason for the PR is because without it, you end up with a rendered file with a single blank line starting the file like this:

```

# Template

Welcome to your new module...
```

Agreed the leading `<%-` isn't required, it's a style I personally use for all code tags in erb templates to prevent superfluous whitespace from being rendered should the block be indented.

PR #214, opened 2019-03-09
Fix needed settings to enable beaker

TheMeier:


README.md
```diff
@@ -220,25 +220,28 @@ following entry to your `.sync.yml` and run `pdk update`.
...
+      - gem: beaker-abs
+        from_env: BEAKER_ABS_VERSION
+        version: '~> 0.1'
+      - gem: beaker-pe
+      - gem: beaker-hostgenerator
+        from_env: BEAKER_HOSTGENERATOR_VERSION
+      - gem: beaker-rspec
+        from_env: BEAKER_RSPEC_VERSION
+.travis.yml:
+  bundler_args: --with system_tests
```
scotje: We should probably add the equivalent settings for Appveyor and Gitlab-CI configs.

TheMeier: @scotje like that?

scotje: Yep! It looks like this has uncovered a bug in the gitlab customization though, I'm still working through that. Also it looks like the config for `.gitlab-ci.yml` has a `beaker` key that we might also want to suggest setting?

TheMeier: @scotje added `beaker: true` I wonder where one could fine a more complete example showing the use of `bekaer.tags` and `beaker.variables`

scotje:
After #244 is merged we can merge this, although I think the `.gitlab-ci.yml` config needs to include the flags from `config_defaults`, I will add a suggestion to the code.

README.md
```diff
@@ -221,25 +221,33 @@ following entry to your `.sync.yml` and run `pdk update`.
...
+.gitlab-ci.yml:
+  bundler_args: --with system_tests
```
scotje:
```suggestion
  bundler_args: --with system_tests --path vendor/bundle --jobs $(nproc)
```

scotje:
Thanks for your contribution @TheMeier !

Issue #215, opened 2019-03-09
Testing should use the right version of ruby for the latest puppet 5 release


ghoneycutt:
Similar to #203 

Latest puppet 5 release (5.5.10) uses ruby 2.4.5

https://puppet.com/docs/puppet/5.5/about_agent.html

DavidS:
If you need (almost) bit-accurate ruby-versions, please use the binaries provided by the PDK packages. Between that and puppet 5 being EOL by the end of month, I'm gonna close this.

PR #216, opened 2019-03-09
(GH-215) Use the correct ruby version for testing Puppet 5

ghoneycutt:
The latest Puppet 5 release is 5.5.10 which uses Ruby 2.4.5.

https://puppet.com/docs/puppet/5.5/about_agent.html


ghoneycutt:
@scotje could you please review

ghoneycutt:
Thanks!

Issue #223, opened 2019-03-22
.travis.yml automatically adds deploy section with incorrect user


ghoneycutt:
The README does not explain how to specify a deploy user/password. This further compounds the issue that pdk-templates adds the deploy section by default as it is not clear how to override the user and specify a password.

Suggest that the deploy section is not automatically added by removing `deploy: true` from `config_defaults.yml`. The `branches` attribute should be left in there. Documentation is needed to show how to add your username/password along with a link to the TravisCI documentation on how to create and encode the password.

rtib:
I think this can be closed, as it was addressed with PR #224 .

rodjek:
@rtib for sure, thanks for catching it! :)

PR #224, opened 2019-03-22
(GH-223) .travis.yml does not deploy by default

ghoneycutt:
Without this patch, the TravisCI configuration will be broken and add a
deployment section with the incorrect password and without documentation
to explain how to do deploys correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 |remove_branches |Allows you to remove default branches set in config_defaults.yml.|
 |notifications   |Allows you to specify the notifications configuration in the .travis.yml file.|
 |remove_notifications   |Allows you to remove default branches set in config_defaults.yml.|
+|user|This string needs to be set to the Puppet Forge user name. To enable deployment the secure key also needs to be set.|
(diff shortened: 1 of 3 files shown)


ghoneycutt:
Fixes #223

rtib:
I'm okay with the idea of having deployment setup optional. But, I think the feature toggle is not really necessary. 

How about this: drop the `deploy` config key and enable the deployment section if `user` and `secure` are set.
This:
- does not require any change for modules having deploy already in place
- reduces `config_defaults.yml`
- simplifies `.sync.yml` for enabling deploy by requiring unavoidable settings only
- keeps the `.travis.yml.erb` template still readable

_BTW, the default setting `user: puppet` favours a single publisher. While I really appreciate Voxpupuli a lot, I think this setting is wrong at pdk-templates._

Look at [10b87b7](https://github.com/rtib/pdk-templates/commit/10b87b78eab0233ab04236ad4126524e41ef2da9).

ghoneycutt:
Good idea, though I don't know how to implement it.

ghoneycutt:
@scotje could you please review

rtib:
> Good idea, though I don't know how to implement it.

[Here is the example](https://github.com/rtib/pdk-templates/commit/ae4c4e6fe64186825a5b1cccb0d389381e45f83b)

PR #228, opened 2019-03-29
Revert "Add Visual Studio Code to gitignote (.vscode/)"

glennsarti:
This reverts commit 8712ac67a4a4132b4a9473042e4f2da930ca9280.

The `.vscode` directory is used to store workspace level overrides, including testing, formatting etc.  It is _designed_ to be checked into repos and should not be ignored by default.

---

The .vscode directory is not required when packaging a module for the forge.
This commit adds this directory to the list of default PDK ignores.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -33,7 +33,6 @@
       - '/update_report.txt'
       - '.DS_Store'
       - '.project'
-      - '.vscode/'
       - '.envrc'
       - '/inventory.yaml'
 .pdkignore:
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -52,6 +52,7 @@
     - '/.travis.yml'
     - '/.yardopts'
     - '/spec/'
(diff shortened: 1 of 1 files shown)


glennsarti:
FYI @ghoneycutt

ghoneycutt:
In which modules have you ever seen anyone check in `.vscode` and it was appropriate?

glennsarti:
Off the top of my head, I have not seen anyone check in `.vscode` in a puppet module, but that doesn't change this PR though.  There are many projects where `.vscode` is checked in appropriately (for example projects that I use/contribute to https://github.com/PowerShell/PowerShellEditorServices, https://github.com/PowerShell/PowerShell and https://github.com/lingua-pupuli/puppet-vscode/tree/master/.vscode).  Also disabling a feature of vscode, as the default, for a PDK user seems quite heavy-handed.

Issue #227, opened 2019-03-22
rubocop complains about default cops


ghoneycutt:
```
Warning: unrecognized cop GetText/DecorateString found in /Users/gh/git/puppet-module-rpcbind/.rubocop.yml
```

Using PDK 1.9.1. You can reproduce with

```
pdk bundle exec rake rubocop
```

which is what happens by default with the TravisCI configs

If you ran `pdk validate` you would not see the warning. Even running with `--debug` this information is not displayed. This is a bug from a UX standpoint as you will see warnings in TravisCI and then not be able to reproduce using the standard tooling ala `pdk validate` instead of resorting to `pdk bundle exec`.

term1n4l:
I followed the recommendations here:

https://github.com/puppetlabs/rubocop-i18n

And resolved this issue

ghoneycutt:
This will be an ongoing issue due to the rapid changes of rubocop. We should adopt a policy of disabling all cops by default and then enabling the ones we want instead of the current policy of taking whatever new thing rubocop thinks is a good idea, having it fail builds and likely not add any real value and then disabling it.

term1n4l:
Just thinking out loud here, could this not be accomplished by dashing out cops in .sync.yml?

scotje:
This should have been fixed by https://github.com/puppetlabs/pdk-templates/pull/280 in PDK 1.14.0, please let us know if it is still an issue.

PR #232, opened 2019-04-01
fix rubocop gettext errors

term1n4l:
as per my comment in #227 this will resolve the gettext error thrown by gitlab-ci and travis-ci


term1n4l:
funnily enough the provider template is what caused the travis-ci build to fail, going to raise that as an issue now...

term1n4l:
this has now been resolved by GH in another pull request, the gemfile might still need updating but that will become apparent soon enough.

PR #234, opened 2019-04-03
(MODULES-8854) Set default changelog_version_tag_pattern to empty

eimlav:
The introduction of this config parameter has broken the changelog
links generated by our changelog rake task. This is because our
modules are tagged using the pure semver i.e. '1.2.3' not 'v1.2.3'.

This change resets this back to the original functionality to ensure
changelogs do not contain broken links. Instead, consumers of the
template should define this parameter in their .sync.yml should they
wish to use it, as is the case with all other changelog related
parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,10 +121,10 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 | Key            | Description   |
 | :------------- |:--------------|
 |requires|A list of hashes with the library to `'require'`, and an optional `'conditional'`.|
-|changelog_user|Sets the github user for the change_log_generator rake task.Optional, if not set it will read the 'author' from the metadata.json file|
-|changelog_project|Sets the github project for the change_log_generator rake task.Optional, if not set it will read the 'name' from the metadata.json file|
-|changelog_since_tag|Sets the github since_tag for the change_log_generator rake task.Required for the changlog rake task|
-|changelog_version_tag_pattern|Template how the version tag is to be generated. Defaults to ```v%s``` which eventually align with tag_pattern property of puppet-blacksmith, thus changelog rake task generated changelog is referring to the correct  |
+|changelog_user|Sets the github user for the change_log_generator rake task. Optional, if not set it will read the 'author' from the metadata.json file|
+|changelog_project|Sets the github project for the change_log_generator rake task. Optional, if not set it will read the 'name' from the metadata.json file|
+|changelog_since_tag|Sets the github since_tag for the change_log_generator rake task. Required for the changlog rake task|
+|changelog_version_tag_pattern|Template how the version tag is to be generated. Optional, if not set it will read the 'version' from the metadata.json file.|
 |default\_disabled\_lint\_checks| Defines any checks that are to be disabled by default when running lint checks. As default we disable the `--relative` lint check, which compares the module layout relative to the module root. _Does affect **.puppet-lint.rc**._ |
 |extra\_disabled\_lint\_checks| Defines any checks that are to be disabled as extras when running lint checks. No defaults are defined for this configuration. _Does affect **.puppet-lint.rc**._ |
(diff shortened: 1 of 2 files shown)


eimlav:
Change originally introduced in #234 for reference.

scotje:
ping @rtib 

Seems reasonable to me to retain the parameter but change the default to empty so that it doesn't affect existing users.

rtib:
Hi @eimlav ,

the default value should be the same as the tag_pattern of puppet-blacksmith. The main intent of #185  was to get a default setting which seamlessly integrate puppet-blacksmith with github-changlog-generator. 

Use cases which change blacksmith's tag_pattern or use any other tooling which end up with tags using a pattern other than 'v%s' may need to adopt the changelog_version_tag_pattern via .sync.yml.

Changing the default to empty will make the result of changelog_future_release become also empty, thus the generated changelog.md will be still broken. If you want the former behaviour back, changelog_version_tag_pattern should be set to `'%s'`. But, that would need set blacksmith's tag_pattern also, which is a bit more complex and will end up in a really odd code like this:
```
Rakefile:
  changelog_version_tag_pattern: '%s'
  extras:
    - |
      Blacksmith::RakeTask.new do |t|
        t.tag_pattern = '%s'
      end
```

I'd prefer not to have that in config_defaults.yml.

scotje:
Hmmm, can we update the Rakefile template (https://github.com/puppetlabs/pdk-templates/blob/master/moduleroot/Rakefile.erb) to ensure that Blacksmith and github-changelog-generator are always configured to use the same pattern, and then have a generalized config__defaults_yml setting that controls both?

scotje:
So we were shipping a mismatched config before where if you tried to use both `github-changelog-generator` and `puppet-blacksmith` the config was mismatched.

Presumably have some users who only use `github-changelog-generator` and some who only use `puppet-blacksmith`, so if we converge the configs and set a default, one of those groups will suddenly experience issues. (In the case of @eimlav and the Puppet modules team, it sounds like they are in the former group.)

I'm inclined to favor the experience of a new user here so having the configs be matched by default seems like the best course of action. @eimlav would the modules team be able to `pdksync` out a change to `.sync.yml` which sets `changelog_version_tag_pattern: '%s'` to resolve this instead of modifying the template default?

rtib:
I'm completely with you @scotje.

The objectives should be:
* consistent setup of all components shipped with newly created modules, i.e. the same tag_pattern for changelog generator and blacksmith
* enable to change settings effectively: experienced users should be able to change any setting, however, I don't want to force them doing so

I was also thinking about a solution of having a single parameter which is generalising the setting of both tools. The implementation could be a bit more complex; #185 was the shortest path and probably just the first step. I'll try to implement that, but it won't go without having a default value.

We need a clever agreement about a default tag_pattern. Especially Puppetlabs and Voxpupuli are crucial here, as both maintain a large amount of modules, while Puppetlabs is tagging mostly with a `'%s'` pattern, Voxpupuli, also maintaining Blacksmith, is using `'v%s'`.

eimlav:
Thanks for the additional background @scotje @rtib . I'm going to chat with the team later in standup and see what they think.

eimlav:
@tphoney @DavidS what are your thoughts on this?

DavidS:
I think Puppet Inc. should be following community standards whenever we are not actively working on changing those standards. 

In this specific example, if puppet-blacksmith defaults to `'v%s'`, I see no reason why that is any worse (or better, for what it's worth) than `'%s'`, and we should be carrying the effort to migrate (or accommodate our special case in the tooling).

eimlav:
Hi all thanks for all the feedback. We have decided to start using the `v%s` pattern going forward so that we are developing our modules in line with the standards that the community expects of us. As such, I'm going to close the PR so thanks for your time :)

PR #235, opened 2019-04-05
Update travis dist to xenial by default

npwalker:
The xenial dist includes ruby 2.4 and 2.5 by default so users can avoid spending 20-30 seconds, per job, installing these versions of ruby just by changing the dist.


npwalker:
@bmjen thoughts?

Issue #134, opened 2018-08-18
Impossible to modify before_script section in .gitlab-ci.yml


Dan33l:
In the template [1], the section `before_script` is only with one variable. And so it can not be overwritten so much. 

[1]  https://github.com/puppetlabs/pdk-templates/blob/master/moduleroot/.gitlab-ci.yml.erb

For instance, i wanted to add the command `pdk --version`. And it looks impossible.

PR #238, opened 2019-04-09
(PDK-1351) Enable custom steps in before_script for GitLab CI

L-Henke:
I have a use case where I need to do additional setup steps for the GitLab pipeline to work (add ca-certificates to allow communication through the proxy). This PR adds the ability to add custom steps in your `.sync.yml` to the before_script in the GitLab CI pipeline.

closes https://github.com/puppetlabs/pdk-templates/issues/134


scotje:
Looks good conceptually, we have an issue with our Travis config for the repo that will block tests but I'll try to get that fixed up shortly.

PR #245, opened 2019-04-30
(PDK-1355) Add hiera_config_ruby to accept ruby snippets

sheenaajay:
Request for review 

Add hiera_execute_config to accept any commands to specify in the spec_helper.rb file

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 | Key            | Description   |
 | :------------- |:--------------|
 |hiera_config|Sets the [`hiera_config`](http://rspec-puppet.com/documentation/configuration/#hiera_config) rspec-puppet parameter.|
+|hiera_config_ruby|Sets the [`hiera_config`](http://rspec-puppet.com/documentation/configuration/#hiera_config) rspec-puppet parameter. A ruby expression returning the path to your hiera.yaml file. `hiera_config` takes precedence if both values `hiera_config` and `hiera_config_ruby` are specified. |
 |mock_with|Defaults to `':mocha'`. Recommended to be set to `':rspec'`, which uses RSpec's built-in mocking library, instead of a third-party one.|
 |spec_overrides|An array of extra lines to add into your `spec_helper.rb`. Can be used as an alternative to `spec_helper_local`|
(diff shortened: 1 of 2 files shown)


moduleroot/spec/spec_helper.rb.erb
```diff
@@ -36,6 +36,9 @@ RSpec.configure do |c|
   <%- if @configs['hiera_config'] -%>
   c.hiera_config = '<%= @configs['hiera_config'] %>'
   <%- end -%>
+  <%- if @configs['hiera_execute_config'] -%>
```
tphoney: should this be an elsif ? is it valid to have both hiera_config and hiera_config ?

tphoney:
there should be documentation updates to explain the new option as well.

moduleroot/spec/spec_helper.rb.erb
sheenaajay: @tphoney make sense. Will make the changes and also update the documentation.

sheenaajay: @tphoney Thank you. Incorporated the review comments and tested the changes.

README.md
```diff
@@ -172,6 +172,7 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 | Key            | Description   |
 | :------------- |:--------------|
 |hiera_config|Sets the [`hiera_config`](http://rspec-puppet.com/documentation/configuration/#hiera_config) rspec-puppet parameter.|
+|hiera_execute_config|Sets the [`hiera_config`](http://rspec-puppet.com/documentation/configuration/#hiera_config) rspec-puppet parameter. It accepts command to find the path to your yaml file.|
```
DavidS: When I read it initially I had a hard time understanding what the difference to `hiera_config` is. Suggestions to improve the situation:

- [ ] use the word `ruby` instead of `execute` to be more specific about the content
- [ ] reword the second sentence to say `a ruby expression returning the path to your hiera.yaml file`` instead of `command ...` to be more specific and helpful.
- [ ] preserve the common prefix by using `hiera_config_ruby` instead of `hiera_ruby_config`. This will make this more searchable. That is folks searching for `hiera_config` will be able to find both
- [ ] specify how this and the regular `hiera_config` values interact. Is it an error if both values are set? Which one takes precedence?

sheenaajay: @DavidS Thanks for the valuable comment.Regarding the last comment.
`hiera_config` takes precedence if both values `hiera_config` and `hiera_config_ruby` are specified.
No errors are thrown if both values are specified.

PR #253, opened 2019-05-29
(FM-8081) Add DeviceShim template too

DavidS:
This missing file was an oversight from #251. It is required for a functioning
transport with puppet.


DavidS:
Part of the implementation for https://github.com/puppetlabs/pdk-planning/pull/31

PR #257, opened 2019-06-05
Override facterdb with default_facts

npwalker:
Prior to this commit, facts provided via default facts would
be ignored if the fact was provided by facterdb.  Which is the
case for any of the standard facts.

After this commit, facts provided via default facts will
override what is provided by facterdb.


npwalker:
The default_facts.yml file says default facts will override rspec-puppet-facts but it doesn't currently.  

https://github.com/puppetlabs/pdk-templates/blob/master/moduleroot/spec/default_facts.yml.erb#L3

PR #260, opened 2019-06-18
Do not use double quotes in the output

vStone:
Rubocop does not like this


vStone:
Looks like this wont be that easy

PR #262, opened 2019-06-27
(PDK-1416) Support Gemfile source option

seanmil:
Add the ability to specify a per-gem source where needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,8 +156,22 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 
 | Key            | Description   |
 | :------------- |:--------------|
-|required|Allows you to specify gems that are required within the Gemfile. Gems can be defined here within groups, for example we use the :development gem group to add in several gems that are relevant to the development of any module.|
+|required|Allows you to specify gems that are required within the Gemfile. Gems can be defined here within groups, for example we use the :development gem group to add in several gems that are relevant to the development of any module and the :system_tests gem group for gems relevant only to acceptance testing.|
 |optional|Allows you to specify additional gems that are required within the Gemfile. This key can be used to further configure the Gemfile through assignment of a value in the .sync.yml file.|
+
+>Within each Gem group defined using the options above one or more gem item definitions may be listed in an array. Each item in that array must be a gem item hash.
+
+| Gem Item Hash Keys | Description  |
+| :----------------- |:-------------|
+|gem|Required option specifying the gem name.|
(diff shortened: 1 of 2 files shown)


scotje:
@seanmil Thanks for the contribution! Would you mind adding something about this option to the README? Once we have that I think this will be ready to merge.

seanmil:
I don't mind adding something about it, but none of the other gem options are documented anywhere in the PDK that I could see. I would have expected them around here: https://github.com/puppetlabs/pdk-templates#gemfile

How would you think that should look? Another table in the Gemfile section or something else?

scotje:
Indeed the existing options don't appear to be documented at all :)

I agree that the section you linked would be the best place for it. I think if we add an example below the existing table and then a new table that lists the options for each entry that would be good. If you want I can work on adding docs for the existing items but if you want to take a pass at it please feel free!

seanmil:
I've added documentation for all of the options in a gem item, an example in the examples area, and clarified a couple of other points around specifying custom gems which I felt could have been a little clearer. Let me know. Thanks!

README.md
```diff
@@ -156,9 +156,24 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
...
-|required|Allows you to specify gems that are required within the Gemfile. Gems can be defined here within groups, for example we use the :development gem group to add in several gems that are relevant to the development of any module.|
+|required|Allows you to specify gems that are required within the Gemfile. Gems can be defined here within groups, for example we use the :development gem group to add in several gems that are relevant to the development of any module and the :system_tests gem group for gems relevant only to acceptance testing.|
 |optional|Allows you to specify additional gems that are required within the Gemfile. This key can be used to further configure the Gemfile through assignment of a value in the .sync.yml file.|
 
+>Within each Gem group defined using the options above one or more gem item definitions may be listed in an array. Each item in that array must be a gem item hash.
+
+| Gem Item Hash Keys | Description  |
+| :----------------- |:-------------|
+|gem|Required option specifying the gem name.|
+|version|Required option to specify version or range of versions required usin RubyGem version syntax.|
```
scotje:
```suggestion
|version|Required option to specify version or range of versions required using [RubyGem version syntax](https://guides.rubygems.org/patterns/#pessimistic-version-constraint).|
```

seanmil: Good catch on the typo and good idea on the link.

scotje:
Thanks for your contribution to PDK @seanmil!

PR #263, opened 2019-06-28
Disable PDK analytics in Travis CI

seanmil:
The recently introduced PDK analytics feature prompts for user
consent on first invocation which is preventing the Travis CI
job from completing successfully on those PDK builds. This
will ensure that PDK analytics are always disabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,15 @@ sudo: false
 dist: trusty
 language: generic
 env:
-- PDK=release
-- PDK=nightly
+  - PDK=release PDK_FRONTEND=noninteractive
+  - PDK=nightly PDK_FRONTEND=noninteractive
 before_install:
-- ".travis/install_pdk.sh"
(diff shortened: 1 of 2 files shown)


scotje:
Hmmm, ENV isn't getting passed through like I expected, going to debug on my own branch.

scotje:
Removed my commit from this PR.

scotje:
@seanmil Figured out how to disable via ENV (and found a PDK bug in the process), what do you think?

scotje:
We could additionally add `PDK_DISABLE_ANALYTICS=true` if we further wanted to ensure no reports were being sent, but ideally we would like to see what people are doing with PDK in CI as well. (https://github.com/puppetlabs/pdk/pull/689 will obviate the need to set ENV to disable the prompt.)

seanmil:
And, for the record, I have no problem with analytics from CI. I just wanted to get it so my other PR would pass checks correctly :)

PR #269, opened 2019-07-16
test against puppet versions of latest PE, latest PE LTS and latest gem

rtib:
Pinning the Puppet versions for spec tests to run tests against:
* latest LTS PE version
* latest PE version
* latest gem version

Versions as of todays [Component versions in recent PE releases](https://puppet.com/docs/pe/2019.0/component_versions_in_recent_pe_releases.html#puppet-enterprise-agent-and-server-components).

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -80,7 +80,7 @@
     - env: PUPPET_GEM_VERSION="5.5.14" CHECK=parallel_spec
       rvm: 2.4.5
       stage: spec
-    - env: PUPPET_GEM_VERSION="6.0.9" CHECK=parallel_spec
+    - env: PUPPET_GEM_VERSION="6.4.2" CHECK=parallel_spec
       rvm: 2.5.3
       stage: spec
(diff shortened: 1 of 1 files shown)


rtib:
While thinking about automating that, is there a official service providing the information needed here? E.g. some ReST-URL responding what's currently contained on [Component versions in recent PE releases](https://puppet.com/docs/pe/2019.1/component_versions_in_recent_pe_releases.html) page?

scotje:
I think we will need a wider consensus that this is a desired set of defaults before we merge this into the base template repo.

On a related note, we are definitely thinking about how to make config changes like this easier to roll out across your organization without having to fork pdk-templates.

pmcmaw:
The IAC team tend to try to avoid pinning to specific puppet versions as it introduces too much maintenance, tracking and updating when newer versions are released. Currently I don't think that the benefits of this specific change outweigh the drawbacks.

pmcmaw:
I am going to close this PR. 

You are welcome to use this configuration on your local setup however we do not think this would be advantageous as a change in pdk-templates.

Many thanks for your contribution
Paula

PR #270, opened 2019-07-17
Add circleci and azure pipelines support

npwalker:


moduleroot/azure-pipelines/spec_job.yml.erb
```diff
@@ -0,0 +1,39 @@
...
+  spec_command: ''
+  ruby_version: ''
+  puppet_gem_version: ''
+
+jobs:
+- job: ${{ parameters.name }}
+  variables:
+    puppet_gem_version: ${{ parameters.puppet_gem_version }}
+  pool:
+    vmImage: 'ubuntu-16.04'
```
glennsarti: This should be parameterised.

moduleroot/azure-pipelines/spec_job.yml.erb
```diff
@@ -0,0 +1,39 @@
...
+      versionSpec: ${{ parameters.ruby_version }}
+      addToPath: true # Optional
+  - script: |
+      bundle -v
+      rm -f Gemfile.lock
+      gem update --system $RUBYGEMS_VERSION --no-document
+    displayName: 'gem update'
+  - script: ruby -v | tee ruby_version.txt
+    displayName: 'save ruby_version file'
+  - task: 1ESLighthouseEng.PipelineArtifactCaching.RestoreAndSaveCacheV1.RestoreAndSaveCache@1
```
glennsarti: Perhaps this whole task should be wrapped in an `if` ?

moduleroot/.circleci/config.yml.erb
```diff
@@ -0,0 +1,83 @@
...
+      puppet_gem_version: << parameters.puppet_gem_version >>
+    steps:
+      - runspecs:
+          spec_command: << parameters.spec_command >>
+
+workflows:
+  version: 2
+  PDK:
+    jobs:
+      - run_specs_on_ruby_puppet_combo:
```
glennsarti: These jobs should be parameterised.  Similar to appveyor and travis

moduleroot/azure-pipelines/spec_job.yml.erb
npwalker: this task only runs if the condition is met.  So it is basically wrapped in an if statement.

glennsarti: Fair enough

moduleroot/.circleci/config.yml.erb
```diff
@@ -0,0 +1,83 @@
...
+    environment:
+      PUPPET_GEM_VERSION: << parameters.puppet_gem_version >>
+
+commands:
+  runspecs:
+    description: "A command for running rake spec"
+    parameters:
+      spec_command:
+        type: string
+        #default: "parallel_spec"
```
glennsarti: nit: remove

scotje:
We should update this so that the config_defaults for these new configs are `deleted` or `unmanaged` so they are opt-in.

We also should note in the PDK release notes that these new configs are community supported.

scotje:
Rebased and added the config_defaults to make the new configs unmanaged by default.

scotje:
Hmmm, the `unmanage` behavior on a new module isn't really what we want I guess, but we don't want it to delete existing configs on an update either... :/ I think we're going to have to punt this to the next release and try to figure out a solution for opt-in templates.

scotje:
This is blocked by PDK-1520 and some further investigation.

npwalker:
What's the concern with including these by default?  All the other CI tool files are included by default and including them provides discoverability.

daianamezdrea:
Hi @npwalker! Given the previous conversation, this PR will be closed. Thank you!

npwalker:
hi @daianamezdrea what's the issue with adding these unmanaged by default?

scotje:
I believe the issue is the current template renderer doesn't apply the "unmanaged" config option when rendering a new module. So the file would be laid down on a new module and then ignored on subsequent `update` operations.

PR #271, opened 2019-07-31
(maint) Add rb-readline to Windows gems

rodjek:
Long term this dependency should be a default gem in the PDK runtimes.


glennsarti:
Light is green, trap is clean.

PR #272, opened 2019-08-02
(PDK-1461) changelog_project fixes

genebean:
Prior to this PR there were two issues:

1. the docs claim that the default value for `changelog_project` comes from the name attribute in `metadata.json`... it actually comes from parsing the `source` attribute.
2. the regex used to parse the `source` attribute to generate the default for `changelog_project` did not account for a url ending in `.git`.

Both of these issues are resolved in this PR.


moduleroot/Rakefile.erb
```diff
@@ -36,7 +36,7 @@ end
 
 def changelog_project
   return unless Rake.application.top_level_tasks.include? "changelog"
-  returnVal = <%= @configs['changelog_project'].inspect -%> || JSON.load(File.read('metadata.json'))['source'].match(%r{.*/([^/]*)})[1]
+  returnVal = <%= @configs['changelog_project'].inspect -%> || JSON.load(File.read('metadata.json'))['source'].match(%r{.*/([^/]*)})[1].reverse.sub(%r{tig\.}, '').reverse
```
scotje:
```suggestion
  returnVal = <%= @configs['changelog_project'].inspect -%> || JSON.load(File.read('metadata.json'))['source'].match(%r{.*\/([^\/]*?)(?:\.git)?\Z})[1]
```

I think something like this would allow you to get rid of the substitutions at the end.

scotje:
Refactored a bit more to make it more resilient to unexpected content in the metadata source key

PR #274, opened 2019-08-21
Adds helper methods for facterdb

logicminds:
Without this change a user doesn't have an easy way to specify 
where their external facterdb facts are.   If you are unfamiliar with external facts
please see the following documentation. 

https://github.com/camptocamp/facterdb#supplying-custom-external-facts

Additionally, this change adds a helper method to toggle the supported operating systems.
This can be handy when testing because sometimes we don't want to see the same test fail
for every supported operating system in the metadata.json. The test_on method allows the user to control which systems are supported without changing the metadata.  This method would be used like so:

```
# spec_helper
def test_on
  @test_on ||= 
   {hardwaremodels: ['x86_64'],
   supported_os: [
       {
           'operatingsystem' => 'Ubuntu',
           'operatingsystemrelease' => ['14.04'],
       },
   ]}
end

# spec/classes/some_class_spec.rb
on_supported_os(test_on).each do |os, os_facts|
    context "on #{os}" do
      it {is_expected.to compile }
    end
end
```

Usage of this method is optional for the user.

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -31,6 +31,38 @@ default_fact_files.each do |f|
   end
 end
 
+# @return [String] - the path to the fixtures directory
+def fixtures_dir
+  @fixtures_dir ||= File.join(__dir__, 'fixtures')
+end
+
+# @return [String] = the path the directory of external facterdb facts
+# Most users will likely not have a custom facterdb_facts git repo defined
+# but if they do then can add it to the fixtures file and name it facterdb_facts.
+# See https://github.com/camptocamp/facterdb#supplying-custom-external-facts
+def mock_facts
+  @mock_facts ||= File.join(fixtures_dir, 'facterdb_facts')
+end
+
+# @return [Hash] - returns a hash of testable operating systems
+# uncomment and replace empty hash
+# modify to your liking, default to all supported
+# operating systems defined in the metadata.json file
+# sometimes you don't want to test on 20 operating systems, this controls that
+def test_on
+  @test_on ||= {}
+  # {hardwaremodels: ['x86_64'],
+  # supported_os: [
+  #     {
+  #         'operatingsystem' => 'Ubuntu',
+  #         'operatingsystemrelease' => ['14.04'],
+  #     },
+  # ]},
+end
+
+# set the mock facts dir for facterdb 
+ENV['FACTERDB_SEARCH_PATHS'] = mock_facts
+
 # read default_facts and merge them over what is provided by facterdb
 default_facts.each do |fact, value|
   add_custom_fact fact, value


scotje:
@rodjek can you take a look at this and see if it makes sense for the August release?

scotje:
Rebase once https://github.com/puppetlabs/pdk-templates/pull/275 is merged should fixup test failures.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -31,6 +31,38 @@ default_fact_files.each do |f|
   end
 end
 
+# @return [String] - the path to the fixtures directory
+def fixtures_dir
```
rodjek: `puppetlabs_spec_helper` already provides this value as `fixture_path` but as a [local variable](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/07978467ed536ce1fcc8b2d0dbb830800e06c9b9/lib/puppetlabs_spec_helper/module_spec_helper.rb#L15). Rather than duplicate this logic, we should change `puppetlabs_spec_helper` to expose this value.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -31,6 +31,38 @@ default_fact_files.each do |f|
...
+def mock_facts
+  @mock_facts ||= File.join(fixtures_dir, 'facterdb_facts')
+end
+
+# @return [Hash] - returns a hash of testable operating systems
+# uncomment and replace empty hash
+# modify to your liking, default to all supported
+# operating systems defined in the metadata.json file
+# sometimes you don't want to test on 20 operating systems, this controls that
+def test_on
```
rodjek: On its own, I don't really see the value in this as the user then has to manually pass it to `on_supported_os`. Instead of this though, we can modify `rspec-puppet-facts` so that `on_supported_os` checks if the test context responds to `test_on` and use that automatically instead of the list from `manifest.json`.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -31,6 +31,38 @@ default_fact_files.each do |f|
...
+# @return [String] - the path to the fixtures directory
+def fixtures_dir
+  @fixtures_dir ||= File.join(__dir__, 'fixtures')
+end
+
+# @return [String] = the path the directory of external facterdb facts
+# Most users will likely not have a custom facterdb_facts git repo defined
+# but if they do then can add it to the fixtures file and name it facterdb_facts.
+# See https://github.com/camptocamp/facterdb#supplying-custom-external-facts
+def mock_facts
```
rodjek: I think this (and the environment variable setting below) would be better suited to be part of `rspec-puppet-facts`.

moduleroot/spec/spec_helper.rb.erb
```diff
@@ -31,6 +31,38 @@ default_fact_files.each do |f|
...
+  # supported_os: [
+  #     {
+  #         'operatingsystem' => 'Ubuntu',
+  #         'operatingsystemrelease' => ['14.04'],
+  #     },
+  # ]},
+end
+
+# set the mock facts dir for facterdb 
+ENV['FACTERDB_SEARCH_PATHS'] = mock_facts
```
rodjek: Rather than override this value completely, we should check if the directory returned by `mock_facts` exists and append it to env var if it does.

moduleroot/spec/spec_helper.rb.erb
logicminds: I like this idea.

moduleroot/spec/spec_helper.rb.erb
logicminds: Seems reasonable.

logicminds:
Closing this as a agree with the comments and will add to the code to the other tools instead.

PR #278, opened 2019-09-18
(maint) Add default ipaddress6 fact

rodjek:
Puppet 6.9.0 now sets a serverip6 fact which is set to the ipaddress6
fact value on the compiling host.

https://github.com/puppetlabs/puppet/commit/48f720a68d9a2903b9ea71561f87af16136d76f6

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -609,6 +609,7 @@ Gemfile:
     # beaker: true
 spec/default_facts.yml:
   ipaddress: "172.16.254.254"
+  ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
   is_pe: false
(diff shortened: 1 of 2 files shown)


config_defaults.yml
```diff
@@ -609,6 +609,7 @@ Gemfile:
     # beaker: true
 spec/default_facts.yml:
   ipaddress: "172.16.254.254"
+  ipaddress6: "FE80:0000:0000:0000:AAAA:AAAA:AAAA"
```
glennsarti: Note - As per https://tools.ietf.org/html/rfc4291#section-2.4  FE80::/10 is link local

Also note that the IP4 address of 172... is not link local but private use.
MAC is a local administered address (i.e. not a vendor specified MAC)

PR #279, opened 2019-09-18
Update the TargetRubyVersion to match the default PDK version

rnelson0:
Validation can fail when using `a: b` format for hashes in rspec tests with `TargetRubyVersion` set to 2.1. Puppet is using at least Ruby version 2.5 these days, the default should be updated.

### DIFF
--- a/moduleroot/.rubocop.yml.erb
+++ b/moduleroot/.rubocop.yml.erb
@@ -7,7 +7,7 @@ defaults = {
   ],
   'AllCops' => {
     'DisplayCopNames' => true,
-    'TargetRubyVersion' => '2.1',
+    'TargetRubyVersion' => '2.5',
     'Include' => [
       './**/*.rb',
     ],


rnelson0:
A more lengthy fix could be to make the version a target for `.sync.yml` options, but I don't think that's truly necessary.

rnelson0:
The test failure doesn't surface that `pdk validate` errors out on rubocop checks, only the return code.

glennsarti:
@rnelson0 This gets VERY complicated very fast.  Apologies if you already know this

Puppet **agent** is  MRI Ruby 2.1, 2.4 and 2.5 (Puppet 4/5/6), HOWEVER, parts of the ruby code (e.g. Types) are executed on the Puppet Server, which is jRuby.

IIRC
Puppet Server 4 uses jRuby 1.7
Puppet Server 5 uses jRuby 1.7 and optional 9K
Puppet Server 6 uses jRuby 9K

The problem is jRuby 1.7 is Ruby 1.9.3 compatible (Thus problems with using the new vs old hash syntax)

For example, we recently hit an issue with frozen strings. It appears (can't prove it yet) that ruby string literals are frozen by default in jRuby 9K, whereas in MRI ruby they are not.

What this meant was module code broke on a PR upgrade from 2018.x to 2019.x and there's no way the MRI ruby version would've ever caught it.

rnelson0:
@glennsarti two things

1) I figured out that my issue was how I was quoting strings (apparently `'/foo': {}` was not valid but failed rubocop, but `/foo: {}` was ok in spite of the leading slash), so that was easier than changing the version. Plus other versions had other issues
2) The `travis/setup.sh` test script does swallow the errors that would have shown why it was failing.

In this case there was definitely a workaround, but I mentioned number 2 above as it might be something to review later.

glennsarti:
Thanks @rnelson0 !

PR #280, opened 2019-09-20
(PDK-1439) Ignore GetText rubocop rules for the off profile

glennsarti:
Previously the rubocop profile builder had not been updated for the i18n rubocop
rules.  This commit:
* Updates the script to work on Windows
* Updates the script for running in the `/rubocop/` directory
* Adds the i18n rubocop gem in the Gemfile so it can be profiled
* Updates the 0.49.1 rubocop profile with the GetText/* rules

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,6 @@
 /tmp/
 /vendor/
 /convert_report.txt
-
+# Rubocop profile builder temporary files
+/rubocop/.rubocop.yml
+/rubocop/.rubocop_todo.yml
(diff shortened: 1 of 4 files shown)


tphoney:
this makes me intensely sad. I understand the reasoning, but i hate seeing work being thrown away.

DavidS:
@TP this only disables it for the `off` profile, which is not used by default in the PDK.

glennsarti:
@DavidS Thanks for that David!  Unfortunately it was more than just that simple change but I got there in the end.

PR #282, opened 2019-09-26
(maint) Add a default CODEOWNERS to new module init.

bmjen:
Adds a CODEOWNERS file to the inital module skeleton generation. This feature
is used internally and is a feature we should promote to facilitate
maintenance.

### DIFF
--- a/moduleroot_init/CODEOWNERS.erb
+++ b/moduleroot_init/CODEOWNERS.erb
@@ -1,0 +1,10 @@
+# You can use a CODEOWNERS file to define individuals or teams that are
+# responsible for code in a repository.
+
+# To learn more about CODEOWNERS, please refer to:
+# https://help.github.com/en/articles/about-code-owners
+
+# Uncomment the line below and add your GitHub Username to start using
+# CODEOWNERS feature.
+
+#* @github_username


bmjen:
cc @lucywyman

lucywyman:
LGTM   Can I merge? DO we need review from PDK team?

bmjen:
@lucywyman Thanks! Will wait for pdk team to review.

glennsarti:
While I'm not against merging a couple of things to consider:

* pdk-templates isn't just for Puppet (the company) modules. Does the broader community want this? Community members have commented previously about Puppet only assuming people use github (as opposed to gitlab, bitbucket etc.)

* This is yet-another-root file (for example have a look at  the puppetlabs/puppetlabs-motd) for our modules which not every community member will want and is something extra to delete.  Is there a way people could opt-in, instead of opt-out?

scotje:
Yeah, this is another case where it would be nice to be able to make things in the `moduleroot_init` opt-in through `pdk config` or similar.

scotje:
Marking as blocked until we implement the ability to opt-in to `moduleroot_init` template files.

glennsarti:
Linking to #295 and Closing.

PR #285, opened 2019-10-18
Ignore *~ files

freiheit:
~ is a common suffix for editor temporary or editor backup files, for editors that aren't vim.


glennsarti:
Thanks for the PR @freiheit !

PR #287, opened 2019-10-31
(PDK-1500) Update Appveyor and Gemfile templates for Litmus

glennsarti:
Previously if a module wanted to use Litmus for acceptance testing it could not
be managed by PDK.  This commit adds the requried boiler plate to the appveyor
configuration file, which then allows Litmus based tasks to be executed.

---

Previously if a module used Litmus for acceptance testing it was possible that
Bundler would choose Development metagems that did not include Litmus which
would then cause CI, or local testing failures.  This commit:

* Adds a `use_litmus` setting to the Gemfile
* When the use_litmus setting is set to true, the Gemfile template is modified
  so that the development gems will be select a version that contains the
  litmus gem

Note that this munging is used because it is not possible to add the complex
logic into the config_defaults file. Nor is it possible to configure this using
knockout prefix in a module's sync.yml.


glennsarti:
Hey @tphoney The Gemfile change is actually for newer Rubies.  While the litmus gem is in the dev metagems, the Gemfile doesn't require it to be used. For example - https://ci.appveyor.com/project/puppetlabs/puppetlabs-concat/builds/28479802/job/acahx073o43kvssh

Notice that it chose `0.3.15` (`Fetching puppet-module-win-dev-r2.5 0.3.15`).  This isn't a build cache or rubgem issue. Bundler just chose that version.

And to be honest, the Gemfile requirement makes sense.

> If the module uses Litmus then the Gemfile should only specify v0.4+ for the Dev metagem, which is where Litmus was introduced.

scotje:
@glennsarti were you hoping to get this into the November release or is a December release milestone OK?

glennsarti:
It can wait.

PR #289, opened 2019-11-13
(maint) Make module deployment in Travis opt-out

glennsarti:
Previously the Travis configuration was setup in a way that made it difficult
to stop auto-forge deployment or use a different tagging system.  This commit
makes the auto deployment of modules optional (opt-out), and allow users to
configure their own tag regex if needed.

This commit preserves the current template defaults, thereby making this a non
breaking change.


glennsarti:
Ref - MODULES-10104

PR #297, opened 2019-12-10
Boolean to remove the before_script section in .gitlab-ci.yml

lukebigum:
Currently the `.gitlab-ci.yml` template always assumes you are using Bundler, and you will always get this section in the file:

```
before_script:
  - bundle -v
  - rm Gemfile.lock || true
  - gem update --system $RUBYGEMS_VERSION
  - gem --version
  - bundle -v
  - bundle install
```

If you need to do a customised Bundler install, or not use bundler at all (eg: Docker image that's already configured), then this section needs to be removed.

I've added a boolean `before_script` to the config of the `.gitlab-ci.yml` file.  Here is an example of how it is used in `.sync.yml`:

```
.gitlab-ci.yml:
  override: true
  custom:
    before_script: false
    custom_jobs:
      pdk:
        script:
          - pdk test unit --parallel
        tags:
          - shell_c7
```

If the boolean is not set, the default behaviour remains the same: the `before_script` section is written as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 
 | Key            | Description   |
 | :------------- |:--------------|
+| os | Set to an array of operating systems to test. See the [TravisCI documentation](https://docs.travis-ci.com/user/multi-os/) for more details. |
 | dist | If specified, it will set the dist attribute. See the [TravisCI documentation](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) for more details. |
 | simplecov      |Set to `true` to enable collecting ruby code coverage.|
 | ruby\_versions  |Define the ruby versions on which you want your builds to be executed.|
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,6 +61,7 @@ Gitlab CI uses a .gitlab-ci.yml file in the root of your repository tell Gitlab
 | rubygems_mirror | Use a custom rubygems mirror url |
 | image          |Define the Docker image to use, when using the Docker runner. Please see the [Using Docker images](https://docs.gitlab.com/ee/ci/docker/using_docker_images.html) docs for specifics.|
 | custom_before_steps |Allows you to pass additional steps to the GitLab CI before_script. Please see the [.gitlab-ci.yml](https://docs.gitlab.com/ce/ci/yaml/#before_script-and-after_script) docs for specifics.|
+| before_script  |If false, removes the default `before_script` section. Useful if you need a customised Bundler install, or to remove Bundler entirely. If the key is unset the default behaviour is to add `before_script`.|
```
scotje: Maybe we should call this setting `default_before_script` so that people aren't confused that they can't specify actual `before_script` contents here?

(Or we could change it so that you _can_ specify a custom `before_script` with this setting and the way to disable the default is to specify this setting with an empty list.)

rodjek:
Whoops, mistyped a push and it automatically closed this PR. Will recreate.

PR #301, opened 2019-12-16
Use force when updating system rubygems

DavidS:
Based on the conversation in https://github.com/rubygems/rubygems/pull/3035
and https://github.com/rubygems/rubygems/issues/3030 this seems to be a
good idea on CI systems where in the cases when this is needed (outdated
base image) we REALLY want the new version to be installed.


DavidS:
Merging this should wait until the linked PRs have landed to make sure we're not missing a change in direction.

scotje:
Doesn't seem like this is needed anymore, but feel free to re-open if you disagree.

Issue #304, opened 2019-12-19
Is it time to upgrade default ruby version in templates?


Felixoid:
## Use Case
The puppet version 4 is far away from the end of life. Maybe, it's proper time to adjust templates to version 5 2.4.5?

## Describe the Solution You Would Like
Add magic comment `# frozen_string_literal: true` to every rb template, update default target version in rubocop and so on

## Describe Alternatives You've Considered
Or it can stay as it is

Felixoid:
I can prepare some PR to start if it's a discussable topic.

kenyon:
#306 should have closed this.

PR #306, opened 2020-01-06
Add header to *.rb files to be ruby 2.4+ compatible

Felixoid:
This PR is a PoC for #304

I would be glad to receive feedback.

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -9,6 +9,8 @@ def requires(item)
   line
 end
 -%>
+# frozen_string_literal: true
+
 require 'puppet_litmus/rake_tasks' if Bundler.rubygems.find_name('puppet_litmus').any?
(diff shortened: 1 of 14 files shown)


scotje:
We are planning to release a PDK 2.x in the coming months which will officially remove the Puppet 4.x (and Ruby 2.1.x) support in PDK, we can certainly use this as a baseline for some of that work but based on our current support lifecycle I don't think we can merge this before 2.x.

scotje:
We could merge all the "frozen string literal" magic comment additions, but we can't change the default Ruby version for Rubocop until PDK 2.0.

Felixoid:
Ok, thanks. I'll be with my laptop on Thursday and can adjust PR then.

Felixoid:
This PR is ready for review.

glennsarti:
Thanks @Felixoid 🎉

PR #309, opened 2020-01-29
Changes for github_changelog_generator Raketask

cdenneen:
- Added github_site for remote server
 - Added github_endpoint for remote server
 - Added gitlab for preparation of gitlab support added in (https://github.com/github-changelog-generator/github-changelog-generator/pull/657)

 These settings will not be added by default. Only if added to `.sync.yml`
 They allow for GitHub Enterprise and Gitlab options down the road.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,9 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 |changelog\_project|Sets the github project name for the change\_log\_generator rake task. Optional, if not set it will parse the `source` from the `metadata.json` file|
 |changelog\_since\_tag|Sets the github since_tag for the change\_log\_generator rake task. Required for the changlog rake task.|
 |changelog\_version\_tag\_pattern|Template how the version tag is to be generated. Defaults to `'v%s'` which eventually align with tag\_pattern property of puppet-blacksmith, thus changelog is referring to the correct version tags and compare URLs. |
+|github_site|Override built-in default for public GitHub. Useful for GitHub Enterprise and other. (Example: `github_site = https://git.domain.tld`) |
+|github_endpoint|Override built-in default for public GitHub. Useful for GitHub Enterprise and other. (Example: `github_endpoint = https://git.domain.tld/api/v4`) |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -149,6 +149,9 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 |changelog\_project|Sets the github project name for the change\_log\_generator rake task. Optional, if not set it will parse the `source` from the `metadata.json` file|
 |changelog\_since\_tag|Sets the github since_tag for the change\_log\_generator rake task. Required for the changlog rake task.|
 |changelog\_version\_tag\_pattern|Template how the version tag is to be generated. Defaults to `'v%s'` which eventually align with tag\_pattern property of puppet-blacksmith, thus changelog is referring to the correct version tags and compare URLs. |
+|github_site|Override built-in default for public GitHub. Useful for GitHub Enterprise and other. |
```
glennsarti: Do you have any examples to add here. For example, would it be `https://xxx.internal.domain` or just `xxx.internal.domain`.  Or a reference to where I'd get more information about what values are valid here.

glennsarti: Also for the other settings below.

cdenneen:
@scotje @glennsarti can you please re-approve. I messed up and ended up having to force-push

moduleroot/Rakefile.erb
```diff
@@ -77,6 +77,15 @@ PuppetLint.configuration.send('<%= option %>')
 if Bundler.rubygems.find_name('github_changelog_generator').any?
   GitHubChangelogGenerator::RakeTask.new :changelog do |config|
     raise "Set CHANGELOG_GITHUB_TOKEN environment variable eg 'export CHANGELOG_GITHUB_TOKEN=valid_token_here'" if Rake.application.top_level_tasks.include? "changelog" and ENV['CHANGELOG_GITHUB_TOKEN'].nil?
+<% if @configs['github_site'] -%>
+    config.github_site = <%= @configs['github_site'].inspect %>
```
glennsarti: Without tests this could be _interesting_.  `.inspect` has bitten us before with YAML files.  But as this is a ruby file and inspect generates ruby code, this is _probably_ safe (?)

PR #310, opened 2020-01-29
Added Gitlab CI Basic only/except options 

cdenneen:
Currently added support for Gitlab CI Basic only/except (https://docs.gitlab.com/ee/ci/yaml/#onlyexcept-basic)

(more work needed for Advanced Support - https://docs.gitlab.com/ee/ci/yaml/#onlyexcept-advanced - as this would have to allow for mixture of arrays and array of hashes)

Example `.sync.yml` configuration:

```yaml
    ruby_versions:
      2.4.9:
        checks:
        - syntax
        - rubocop
        - parallel_spec
        allow_failure:
        - rubocop
        except:
          syntax:
          - tags
          - master
          rubocop:
          - tags
          - master
        only:
          parallel_spec:
          - tags
          - master
```


moduleroot/.gitlab-ci.yml.erb
```diff
@@ -91,6 +91,22 @@ before_script:
   allow_failure: true
 <%         end -%>
 <%       end -%>
+<%       if options['except'] -%>
+<%         if options['except'].has_key?(check) -%>
+  except:
+<%           options['except'][check].each do |ref| -%>
```
scotje: This might need to guard against `options['except'][check]` being `nil` since `has_key?` won't protect from the value being `nil`.

moduleroot/.gitlab-ci.yml.erb
```diff
@@ -91,6 +91,22 @@ before_script:
...
+    - <%= ref %>
+<%           end -%>
+<%         end -%>
+<%       end -%>
+<%       if options['only'] -%>
+<%         if options['only'].has_key?(check) -%>
+  only:
+<%           options['only'][check].each do |ref| -%>
```
scotje: Same as above. (This might need to guard against `options['only'][check]` being `nil` since `has_key?` won't protect from the value being `nil`.)

moduleroot/.gitlab-ci.yml.erb
scotje:
```suggestion
<%           options['except'][check] && options['except'][check].each do |ref| -%>
```

moduleroot/.gitlab-ci.yml.erb
scotje:
```suggestion
<%           options['only'][check] && options['only'][check].each do |ref| -%>
```

moduleroot/.gitlab-ci.yml.erb
```diff
@@ -91,6 +91,22 @@ before_script:
   allow_failure: true
 <%         end -%>
 <%       end -%>
+<%       if options['except'] -%>
```
glennsarti: This could be merged into a single `if` instead of nested e.g.

```
<% unless options['except'].nil? || options['except']['check'].nil? -%>
    except:
<% options['except'][check].each do |ref| -%>
   - <%= ref %>
<%  end -%>
<%  end -%>
```

moduleroot/.gitlab-ci.yml.erb
```diff
@@ -91,6 +91,22 @@ before_script:
...
+<%         if options['except'].has_key?(check) -%>
+  except:
+<%           options['except'][check] && options['except'][check].each do |ref| -%>
+    - <%= ref %>
+<%           end -%>
+<%         end -%>
+<%       end -%>
+<%       if options['only'] -%>
```
glennsarti: This could be merged into a single `if` instead of nested e.g.

```
<% unless options['only'].nil? || options['only']['check'].nil? -%>
    except:
<% options['only'][check].each do |ref| -%>
   - <%= ref %>
<%  end -%>
<%  end -%>
```

README.md
```diff
@@ -56,7 +56,7 @@ Gitlab CI uses a .gitlab-ci.yml file in the root of your repository tell Gitlab
...
 | bundler\_args   |Define any arguments you want to pass through to bundler. The default is `--without system_tests --path vendor/bundle --jobs $(nproc)` which avoids installing unnessesary gems while installing them to the `vendor/bundler.|
-| ruby_versions  |Define a list of ruby_versions to test against. Each version can have a series of sub-keys that are options. `checks` is the rake command(s) to run during the job. `puppet_version` sets the PUPPET_GEM_VERSION environment variable. `allow_failure` is an array of `checks` where you want to allow failures. `tags` is an array of Gitlab CI Runner tags.
+| ruby_versions  |Define a list of ruby_versions to test against. Each version can have a series of sub-keys that are options. `checks` is the rake command(s) to run during the job. `puppet_version` sets the PUPPET_GEM_VERSION environment variable. `allow_failure` is an array of `checks` where you want to allow failures. `tags` is an array of Gitlab CI Runner tags. `except`/`only` is an hash of `checks` with array of references of conditions for the `checks`.
```
glennsarti: There's a lot of information within that comment now.  Perhaps it's time to split it out like the `deploy_to_forge` or `litmus` settings in Travis.

Also adding examples is _really_ useful.

cdenneen:
@scotje @glennsarti can you please re-approve. I messed up and ended up having to force-push

scotje:
@cdenneen it looks like this PR picked up the changes from the other PR as well

scotje:
@cdenneen `options['except'][check].nil?` will handle checking both that the key exists and that it's value is not `nil`. (Asking a Hash for a key that doesn't exist returns `nil` which is the same as if the key did exist but it's value was set to `nil`.)

PR #311, opened 2020-01-30
Add `default_facter_version` parameter to `.sync.yml`

rnelson0:
Currently, the pdk is pinned to facterdb 0.8.2 which doesn't auto-detect facter versions as well as we could wish, but users can override the facter version. This new parameter in the sync file allows the setting to persist across a `pdk update`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ Travis uses a .travis.yml file in the root of your repository to learn about you
 
 | Key            | Description   |
 | :------------- |:--------------|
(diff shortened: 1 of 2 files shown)


DavidS:
What would it take to update the facterdb version instead? Would that solve your problem?

rnelson0:
I think this is in part because puppet's facterdb dependency is ~>0.8.1 so it can't go to 1.x, which does better at auto-detecting which version of facts to use, and that should definitely be fixed. But this would also allow people to pin to the version of facts, if desired.

PR #320, opened 2020-03-16
(PDK-1591) Add option to enable legacy_facts puppet-lint plugin

rodjek:
Keen observers may note that only one template is affected by this change in contradiction to the description of the `common` namespace introduced here. In a future release I plan to add the ability to disable legacy facts during rspec-puppet catalogue testing which will also be controlled by this setting.


DavidS:
Since when is `config_for` available?

rodjek:
> Since when is `config_for` available?

I snuck it in a while ago when I needed a way to build the `.puppet-lint.rc` template using options defined for the `Rakefile` template. :)

PR #324, opened 2020-04-06
(maint) - Quick fix to Layout/LineLength

david22swan:


DavidS:
On the change in https://github.com/puppetlabs/puppetlabs-concat/pull/639 rubocop says:

```
/home/travis/build/puppetlabs/puppetlabs-concat/.rubocop.yml: Layout/LineLength has the wrong namespace - should be Metrics
```
https://travis-ci.org/github/puppetlabs/puppetlabs-concat/jobs/670690913#L607

should we really commit this?

sheenaajay:
@DavidS Ohh noo...I think the fix only for the latest ruby version and not for 2.5.3 which we use on all our modules. Will close the PRs and pdk templates PR. Thanks @david22swan

PR #332, opened 2020-05-19
Add global_variables key for GitLab CI to set defaults that can be overridden by jobs

cdenneen:
This would incorporate global variables in the job level variables.
This could probably be even turned into some sort of YAML anchor instead

### DIFF
--- a/moduleroot/.gitlab-ci.yml.erb
+++ b/moduleroot/.gitlab-ci.yml.erb
@@ -86,6 +86,11 @@ before_script:
 <%       if options['rubygems_version'] -%>
     RUBYGEMS_VERSION: '<%= options['rubygems_version'] -%>'
 <%       end -%>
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
@cdenneen any luck following up on @DavidS's review from May 20? It's been about a month, we typically look to close stale PRs after about 90 days of inactivity.

cdenneen:
> I'm fine with not using YAML anchors as the rendering and reading is already complex enough.
> 
> Please do add the new key to the README, potentially nabbing the description from .gitlab-ci.yml (and verifying that it actually works similarly).

There is no key. There is already a global_variables key the problem is any custom job setting a variables block will override all global_variables so this allows them to carry forward.

DavidS (approved):
I'm sorry, I originally mis-read this as "global" being top-level to the default_configs.yml. This is really only internal to the gitlab configuration.

PR #337, opened 2020-07-03
Conditional can't be used with false. Since default is true it is not needed

cdenneen:
Fix previous change where false value would not work because of conditional.


daianamezdrea:
Hi @cdenneen, this looks good ! Thank you !

PR #341, opened 2020-07-10
Add starter Forge publish GitHub action

binford2k:
Adds a starter GitHub action. Users will have to enable it in their `sync.yml` or local template fork for now.

### DIFF
--- a/moduleroot/.github/workflows/publish.yaml.erb
+++ b/moduleroot/.github/workflows/publish.yaml.erb
@@ -1,3 +1,4 @@
+<% if configs['enabled'] -%>
 name: Build and publish to Puppet Forge
 
 on:
@@ -8,7 +9,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    if: ${{ false }} # remove this line and create a FORGE_API_KEY secret to enable the job
     steps:
     - name: Get latest tag
       run: echo ::set-env name=RELEASE_VERSION::$(echo ${GITHUB_REF:10})
@@ -21,3 +21,10 @@ jobs:
       env:
        FORGE_API_KEY: ${{ secrets.FORGE_API_KEY }}
        REPOSITORY_URL: https://forgeapi.puppet.com/v3/releases
+<% else -%>
+# Add the following key to your .sync.yml and create a FORGE_API_KEY GitHub
+# secret to enable this job.
+#
+# .github/workflows/publish.yaml:
+#   enabled: true
+<% end -%>
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -8,6 +8,8 @@ common:
     '*.pp': 'eol=lf'
     '*.sh': 'eol=lf'
     '*.epp': 'eol=lf'
+.github/workflows/publish.yaml:
+  enabled: false # requires a FORGE_API_KEY GitHub secret
 .gitignore:
   required: &ignorepaths
       - '.git/'
--- a/moduleroot/.github/workflows/publish.yaml
+++ b/moduleroot/.github/workflows/publish.yaml
@@ -1,0 +1,23 @@
+name: Build and publish to Puppet Forge
+
+on:
+ push:
+  tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: ${{ false }} # remove this line and create a FORGE_API_KEY secret to enable the job
+    steps:
+    - name: Get latest tag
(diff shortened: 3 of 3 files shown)


moduleroot/.github/workflows/publish.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+      - v[0-9]+.[0-9]+.[0-9]+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Get latest tag
+      run: echo ::set-env name=RELEASE_VERSION::$(echo ${GITHUB_REF:10})
+    - name: Clone repository
+      uses: actions/checkout@v1
```
jpogran, line 16: According to [github actions docs](https://github.com/actions/checkout#usage) the checkout action ref is automatically set to the reference or SHA for the event that triggered the action. I _think_ that means that its that tag, so the 'Get latest tag' step might be unnecessary

moduleroot/.github/workflows/publish.yaml.erb
```diff
@@ -0,0 +1,30 @@
...
+ push:
+  tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Get latest tag
+      run: echo ::set-env name=RELEASE_VERSION::$(echo ${GITHUB_REF:10})
```
jpogran, line 14: If the checkout action does actually need the ref, then you could use github actions `set-output` instead like I do in (vscode)[https://github.com/puppetlabs/puppet-vscode/blob/1ec3cb73d3a6cf1675c35cb1e91fdf15ba2fa23c/.github/workflows/release.yml#L37] . That creates a variable that you can reference inside the [yaml](https://github.com/puppetlabs/puppet-vscode/blob/1ec3cb73d3a6cf1675c35cb1e91fdf15ba2fa23c/.github/workflows/release.yml#L57) instead of relying environment variables.

This might not be an improvement per say, and instead might be yak shaving. Also, beware I found `set-output` only works in bash.

daianamezdrea:
Hi @binford2k, I would like to merge this PR, did you see anything from @jpogran suggestions that you can use for improvement? Thank you, cheers!

moduleroot/.github/workflows/publish.yaml.erb
binford2k: I think you're correct. I just copied the example from that action's README, but the code actually grabs the release version from metadata, not the tag. https://github.com/barnumbirr/action-forge-publish/blob/eb321624ad542407801ff22cbf58426e4ed616e0/entrypoint.sh#L7

I have a few improvements to make to `node_encrypt`. Maybe I'll get some time to work on that this weekend and test publishing with this step removed.

nmaludy:
@binford2k FYI i've been working on converting some of our Travis stuff over to GitHub actions example: https://github.com/EncoreTechnologies/puppet-powershellmodule/blob/master/.github/workflows/build.yaml

I'm interested in helping this along. I think the only piece missing from my version is the following:
- acceptance testing 
- templatizing and turning hard coded things into variables and options

DavidS (changes requested):
I really like @nmaludy 's [workflow](https://github.com/EncoreTechnologies/puppet-powershellmodule/blob/master/.github/workflows/build.yaml) to complement our existing travis/appveyor configurations.

Like I said [elsewhere](https://github.com/puppetlabs/pdk-templates/pull/354#pullrequestreview-526041853) we should tag this and other optional enhancements as `unmanaged: true` in the default configs so we do not overwhelm newcomers.

I've changed this PR to draft status to re-consider whether we don't just want to take over @nmaludy 's workflow straight up.

Related, we're also working on getting acceptance testing running on github actions. Please see https://github.com/puppetlabs/puppetlabs-motd/tree/main/.github/workflows for our first steps.

sanfrancrisko:
Given we now have the [`release`](https://github.com/puppetlabs/pdk-templates/blob/47fb99088009dbc070637b3f6cbf1ca4b087ae82/moduleroot/.github/workflows/release.yml.erb) from @carabasdaniel merged, I'm going to close this PR, as I think most of what we need is functionally in that workflow job.

If there are additions or things from your workflow, @nmaludy, that we still don't have but think are beneficial, we could enhance the [`release`](https://github.com/puppetlabs/pdk-templates/blob/47fb99088009dbc070637b3f6cbf1ca4b087ae82/moduleroot/.github/workflows/release.yml.erb) workflow in a new PR?

PR #343, opened 2020-08-04
(IAC-1026) - Update default travis/appveyor branches from master to main

david22swan:


raphink:
I know it's a bit late (and I wish I had seen this before getting bitten by it today), but breaking compatibility for hundreds of existing repos could have been avoided by supporting both `main` and `master` branches.

PR #347, opened 2020-08-26
(maint) - Update `main` with PR merged into `master`

david22swan:


scotje:
@david22swan +1 to the rename, were you planning to update the various CI bits that will need to be updated as well or should I take a look at that?

scotje:
This looks like it has broken the packaging step in CI: https://jenkins-platform.delivery.puppetlabs.net/job/platform_pdk_pdk-van-build_master/BUILD_TARGET=debian-9-amd64,SLAVE_LABEL=beaker/1117/console

david22swan:
@scotje Hey, sorry for the late reply, missed the notification on this. Didn't realize there would be such a knock-on effect from this, slightly confused on where the error is coming from as the master branch has not yet been removed.
Could you point me in the direction of the changes that need to be made please?

PR #349, opened 2020-09-21
(#348) Add .devcontainer for vscode

### LINKED ISSUE #348: Add .devcontainer for vscode
## Use Case
Opening a PDK project in a vscode development container creates a top-level `.devcontainer` that is not managed by PDK and is not ignored by git.

## Describe the Solution You Would Like
Ideally the PDK templates should include a `.devcontainer` that matches what is included in the PDK development container.

## Describe Alternatives You've Considered
Presumably we don't want to ignore the directory.  The only other alternative is to leave it unmanaged and force the user to add the file manually.

## Additional Context
https://code.visualstudio.com/docs/remote/create-dev-container

silug:
Fixes #348.

### DIFF
--- a/moduleroot/.devcontainer/Dockerfile
+++ b/moduleroot/.devcontainer/Dockerfile
@@ -1,0 +1,6 @@
+FROM puppet/pdk:latest
+
+# [Optional] Uncomment this section to install additional packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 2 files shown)


jpogran (approved):

michaeltlombardi (approved):
I'm way into this, would absolutely use devcontainers if they were available. 🎉 this is awesome!

binford2k (approved):

PR #354, opened 2020-10-17
Add gitpod support for puppet modules

logicminds:
* This adds initial support for gitpod puppet development.  Any user
    who opens the module in gitpod would automatically
    stand up a PDK and puppet development environment.

    While this is a first stab at this file I expect further customizations
    to follow and allow the user to utilize the sync or config_defaults
    to tweak their own unique gitpod.

    I decided to leave the clickable banner that normally goes in the readme out of this changeset as I don't know
    what the user's gitpod link would actually be.  I am sure a config variable might help with this but decided to KISS.

An example of this setup can be viewed here by clicking this link: [Gitpod ready-to-code](https://gitpod.io/#https://gitlab.com/nwops/control-repo.git)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,17 @@ Gitlab CI uses a .gitlab-ci.yml file in the root of your repository tell Gitlab 
 | custom_before_steps |Allows you to pass additional steps to the GitLab CI before_script. Please see the [.gitlab-ci.yml](https://docs.gitlab.com/ce/ci/yaml/#before_script-and-after_script) docs for specifics.|
 | default_before_script  |If false, removes the default `before_script` section. Useful if you need a customised Bundler install, or to remove Bundler entirely. If the key is unset the default behaviour is to add `before_script`.|
 
+### Gitpod configuration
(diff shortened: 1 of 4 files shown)


michaeltlombardi:
We're discussing this one internally a bit, as it will effect _all_ Puppet modules. :)

logicminds:
Yes similar to Travis, appveyor and other stuff people ignore if they don't use.

DavidS (changes requested):
I'm happy to merge this, if we tag this as `unmanaged: true` in the config defaults. That way the functionality is there for people who want it.

> other stuff people ignore if they don't use
We've heard enough complaints about the confusing maze of files we deploy for newcomers that I don't want to add on top of that.
I'd disable all non-essential stuff too for a pdk 2.0 release.

@binford2k once we get this merged, I suggest we enable it on all supported modules to see if that brings in more/better contributions.

logicminds:
@DavidS made the changes as requested and updated the README as well to show how to enable.

DavidS (approved):

PR #356, opened 2020-10-20
(#353) Add an empty, commented .sync.yml

### LINKED ISSUE #353: Include an empty, commented .sync.yml
## Use Case
For new PDK users, figuring out how to customize a module with `.sync.yml` is difficult.

## Describe the Solution You Would Like
pdk-templates should include an example `.sync.yml` in `moduleroot_init` for documentation purposes.

silug:
Fixes #353.


david22swan (approved):
Seem's like a good idea and I can't see any problems

david22swan:
@silug Thank's for raising this. Look's like a good improvement for the future.

PR #359, opened 2020-11-10
Default new modules to rspec-mocks

DavidS:
For the longest time we've been recommending rspec-mocks as the default mocking framework, but haven't been able to break backwards compatibility for modules using both rspec-mocks and mocha together. Now, at least for new modules we can do better.

### DIFF
--- a/moduleroot_init/.sync.yml.erb
+++ b/moduleroot_init/.sync.yml.erb
@@ -5,4 +5,6 @@
 #
 # See https://github.com/puppetlabs/pdk-templates/blob/main/config_defaults.yml
 # for the default values.
---- {}
+---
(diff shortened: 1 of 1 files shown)


DavidS:
closing and re-opening to tickle travis

scotje:
I think people will still have to `pdk update` after generating a new module to get `rspec-mocks` configured, right?

DavidS:
@scotje ouch, you're right - this'll need more work on the pdk side to have the defaults available before everything is rendered out :-(

logicminds:
Awesome. I was complaining last week about this.  Thanks for doing this.

DavidS:
Fixed in #404

PR #364, opened 2020-11-23
Avoid second failure if provision doesn't create inventory

DavidS:
Without an inventory this would never be able to create a
correct tear_down call. Instead it is causing a second error
that is confusing the github actions log viewer. This change
avoids running the `litmus:tear_down` rake task when provisioning
has failed.


scotje (approved):
Seems reasonable!

PR #367, opened 2020-11-25
github actions: Don't retry at the workflow level

DavidS:
Rely on litmus' provisioning retry loop instead


DavidS:
see https://github.com/puppetlabs/puppetlabs-testing/pull/323 for a test run

sheenaajay (approved):

PR #370, opened 2020-12-08
(SEC-250) Pin non-github provided actions to a specific SHA

DavidS:
> Malicious code can be inserted into any GitHub action, even those which are tagged.
-- https://julienrenaux.fr/2019/12/20/github-actions-security-risk/

This is mitigated through using https://github.com/mheap/pin-github-action to update non github-provided actions to use a full SHA instead of relying on mutable tags or branch names.


carabasdaniel (approved):
👍

carabasdaniel:
Merging PR as Travis CI failure is not related to these changes

PR #377, opened 2020-12-18
Update how we handle at-risk action sources

DavidS:
## Security Considerations on Github Actions

As explained in [Use GitHub actions at your own risk](https://julienrenaux.fr/2019/12/20/github-actions-security-risk/),
when running github actions from outside the organisation,
there is a risk that symbolic references get taken over by malicious actors.
Similar things happened before in other ecosystems and other packaging registries.
The blog post goes on to suggest pinning to specific SHAs and provides some tooling to do so.
The downsides for us are that the tooling doesn't work well with our ERB templating,
and the additional cost of updating the SHAs across all modules.
Instead we fork at-risk actions into the puppetlabs namespace and use them from there.
This allows us to consume updates at our pace and deploy changes across all modules without delay,
while avoiding actions that surreptitiously change while we're not looking.

Since this still has some overhead, we exclude some "big-name" action maintainers:
* Anything maintained by Github, e.g. [https://github.com/actions](https://github.com/actions)
* Anything maintained as part of a bigger OSS project we're using, like [https://github.com/ruby/setup-ruby](https://github.com/ruby/setup-ruby)
* Anything maintained by a Puppet employee

### Updating actions guitelines

To keep efforts low when updating actions, we list all forked actions here.
To keep confusion to a minimum, the version we use is always on a `pdk-templates` branch.
This way we can update (`git fetch`/`git push`) forked repositories with no prejudice, test out the changes, and only then update the `pdk-templates` branch.

* [kvrhdn/gha-buildevents](https://github.com/kvrhdn/gha-buildevents) ➡️ [puppetlabs/kvrhdn-gha-buildevents](https://github.com/puppetlabs/kvrhdn-gha-buildevents/tree/pdk-templates)
* [Gamesight/slack-workflow-status](https://github.com/Gamesight/slack-workflow-status) ➡️ [puppetlabs/Gamesight-slack-workflow-status](https://github.com/puppetlabs/Gamesight-slack-workflow-status/tree/pdk-templates)

The repos have restricted access only to [@modules](https://github.com/orgs/puppetlabs/teams/modules) team members.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,10 +353,11 @@ Since this still has some overhead, we exclude some "big-name" action maintainer
 ### Updating actions guitelines
 
 To keep efforts low when updating actions, we list all forked actions here.
-To keep confusion to a minimum, the version we use is always on a `pdk-templates` branch.
-This way we can update (`git fetch`/`git push`) forked repositories with no prejudice, test out the changes, and only then update the `pdk-templates` branch.
+To keep confusion to a minimum, the version we use is always on a `pdk-templates-v1` branch.
+This way we can update (`git fetch`/`git push`) forked repositories with no prejudice, test out the changes, and only then update the `pdk-templates-v1` branch.
+If we later need to support multiple versions of an action as we roll out changes, we can increment the `-v1` part in the branch name to manage multiple versions.
 
-* [kvrhdn/gha-buildevents](https://github.com/kvrhdn/gha-buildevents) ➡️ [puppetlabs/kvrhdn-gha-buildevents](https://github.com/puppetlabs/kvrhdn-gha-buildevents/tree/pdk-templates)
-* [Gamesight/slack-workflow-status](https://github.com/Gamesight/slack-workflow-status) ➡️ [puppetlabs/Gamesight-slack-workflow-status](https://github.com/puppetlabs/Gamesight-slack-workflow-status/tree/pdk-templates)
+* [kvrhdn/gha-buildevents](https://github.com/kvrhdn/gha-buildevents) ➡️ [puppetlabs/kvrhdn-gha-buildevents](https://github.com/puppetlabs/kvrhdn-gha-buildevents/tree/pdk-templates-v1)
+* [Gamesight/slack-workflow-status](https://github.com/Gamesight/slack-workflow-status) ➡️ [puppetlabs/Gamesight-slack-workflow-status](https://github.com/puppetlabs/Gamesight-slack-workflow-status/tree/pdk-templates-v1)
 
 The repos have restricted access only to [@modules](https://github.com/orgs/puppetlabs/teams/modules) team members.
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ If you are using Gitpod you will need to opt-in and enable gitpod support for pd
 .gitpod.Dockerfile:
   unmanaged: false
 .gitpod.yml:
-  unmanaged: false  
+  unmanaged: false
 
 ```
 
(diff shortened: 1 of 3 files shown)


DavidS:
https://github.com/puppetlabs/puppetlabs-testing/pull/329 has passed green from this change

ekohl:
I think these are valid concerns. In Voxpupuli we only rely on `actions/checkout` and `ruby/setup-ruby` and trust those. I've tried to keep everything minimal. My guidelines are to only perform things you can execute yourself in a shell (rake tasks) and limit it to what needs to happen (git checkout, ensure env has the needed tools). So far I haven't really missed instrumentation like honeycomb or report to Slack (I'm not even on Slack, but not sure I would look at them in IRC) though I really wish there was a better way to report test suites (https://github.com/Drieam/rspec-github is on my list to explore).

Forking sounds like a decent solution but using a branch named `pdk-templates` means you can't really do major versions. I'd advise to at least use some `v1` tag in there.

DavidS:
> Forking sounds like a decent solution but using a branch named pdk-templates means you can't really do major versions. I'd advise to at least use some v1 tag in there.

That's a good catch, I'll update that.

README.md
```diff
@@ -353,10 +353,11 @@ Since this still has some overhead, we exclude some "big-name" action maintainer
 ### Updating actions guitelines
 
 To keep efforts low when updating actions, we list all forked actions here.
-To keep confusion to a minimum, the version we use is always on a `pdk-templates` branch.
-This way we can update (`git fetch`/`git push`) forked repositories with no prejudice, test out the changes, and only then update the `pdk-templates` branch.
+To keep confusion to a minimum, the version we use is always on a `pdk-templates-v1` branch.
```
ekohl, line 356: Just thinking out loud: should it follow upstream? If they are on v1, it should be v1. If they're on v2 but v1 was never deployed in pdk-templates then it's confusing when they don't match.

DavidS: Hmmm - I was thinking this is the template's version, as that's what we control here. Not sure there is one good solution.

ekohl: My reasoning is that I'd want to treat it more as a mirror than a fork so the fewer changes compared to upstream, the better. As a community member, having to rely on Puppetlabs has proven to block us from time to time. Making it easier to use upstream actions again is IMHO a good thing.

DavidS: Would that blocking concern be addressed by opening these repos up to Trusted Contributors?

DavidS: Let me try to phrase that clearer:

Would that blocking concern be addressed by giving you (and others) write access to these repositories through a Trusted Contributor Programme?

ekohl: I'm not sure. I don't use the pdk-templates so I would have no incentive to work on that.

However, I do think that you should have a rule that any modifications should be sent to the upstream action and you only merge/cherry pick from upstream. Only in exceptional areas would you want it to become a real fork. The project's README should describe that.

DavidS: > As a community member, having to rely on Puppetlabs has proven to block us from time to time. 
> I'm not sure. I don't use the pdk-templates so I would have no incentive to work on that.
I'm getting mixed messages here.

> However, I do think that you should have a rule that any modifications should be sent to the upstream action and you only merge/cherry pick from upstream. Only in exceptional areas would you want it to become a real fork. The project's README should describe that.

Agreed - there should be no changes on the `pdk-templates-*` branches that are not (at least in a PR) upstream.

ekohl: > > As a community member, having to rely on Puppetlabs has proven to block us from time to time.
> > I'm not sure. I don't use the pdk-templates so I would have no incentive to work on that.
> 
> I'm getting mixed messages here.

I'm saying that as a community member I've experienced in the past that it can be annoying to be blocked on something. In my experience, Puppet (still have to get used to not call it Puppetlabs) is slow to review PRs. Just of the top of my head, https://github.com/puppetlabs/puppetlabs_spec_helper/pull/327 has been open for a month without even a review.

Now I'm experienced in being a maintainer so I know it can be hard to deal with. What I am suggesting that the process should be designed in a way that with minimal friction users can work around it. Being as close as possible to upstream, also encourages the best open source behavior for the largest amount of people.

Does that clarify it?

DavidS: > Does that clarify it?

yes, then we're on the same page what we're talking about.

Again, would **all those concerns** be addressed by giving you (and others) write access to **all these repositories** through a Trusted Contributor Programme? Not just in this particular case, or that PR you're referencing; in the most general sense of "would the puppet community use a more cohesive toolset and move forward quicker if y'all didn't need waiting on puppet employees to push a green button for you?"

ekohl: I think in general more contributors does help. Of course there's never a guarantee, but it's the right way to deal with it IMHO.

> if y'all didn't need waiting on puppet employees to push a green button for you?"

Note that often actual releases are needed. Often you need a gem or module release for production cases. I know this can also be PRs, but I'm being explicit here.

ekohl (approved):

README.md
ekohl: I should also mention that I refuse to use the PDK since it contains telemetry and fundamentally disagree with that. Having an opt-out is IMHO irrelevant. Such a thing is unacceptable to me and by using the opt-out, you would legitimate the tool.

DavidS: I've added verbiage relating to limiting modifications to the README and screencapped your comment on trusted contributor rights for product mgmt.

sanfrancrisko (approved):

PR #380, opened 2020-12-27
Update README.md

jkroepke:
Version 1.0 of puppet-module-posix-system does not require breaker anymore.


tphoney:
Hi @jkroepke thanks for the submission however we have removed that section of the readme, apologies for the change in flight.

PR #383, opened 2021-01-05
Fix handling of `pending` cops; configure some egregious cops

DavidS:


tphoney (approved):
I have tried this out on a few modules. it looks good to me.

PR #385, opened 2021-01-05
(MAINT) Fix bundler resolution issues for older puppet versions

scotje:
The way bundler resolves the puppet_litmus dependencies is currently
preventing dependency resolution when Puppet version is less than
6.11.0. This updates the PDK-generated Gemfile to fall back to the
pre-1.0 releases of the `puppet-module-*-system` gems when the Puppet
version is less than 6.11.0.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -592,9 +592,22 @@ Gemfile:
     ':system_tests':
       - gem: 'puppet-module-posix-system-r#{minor_version}'
         version: '~> 1.0'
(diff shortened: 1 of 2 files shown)


DavidS:
Hmm ... we could drop the system testing stuff completely on non-current puppet versions since the gem-versions are only used to run litmus/bolt and not for litmus-testing anyways.

scotje:
Identified the bug in PDK that is causing tests to fail but this is a chicken/egg problem in terms of getting all the tests to pass so going to merge this first.

PR #386, opened 2021-01-08
(MAINT) Remove puppet-module-system gems for older Puppet versions

scotje:
Following @DavidS's suggestion, I think this is simpler than trying to resolve all the dependency conflicts...

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -590,20 +590,9 @@ Gemfile:
         version: '~> 1.0'
         condition: "Gem::Requirement.create(['>= 6.11.0', '< 8.0.0']).satisfied_by?(Gem::Version.new(puppet_version.dup))"
         platforms: ruby
-      - gem: 'puppet-module-posix-system-r#{minor_version}'
-        version: '~> 0.4'
-        condition: "Gem::Requirement.create('< 6.11.0').satisfied_by?(Gem::Version.new(puppet_version.dup))"
-        platforms: ruby
       - gem: 'puppet-module-win-system-r#{minor_version}'
         version: '~> 1.0'
         condition: "Gem::Requirement.create(['>= 6.11.0', '< 8.0.0']).satisfied_by?(Gem::Version.new(puppet_version.dup))"
-        platforms:
-          - mswin
-          - mingw
-          - x64_mingw
-      - gem: 'puppet-module-win-system-r#{minor_version}'
-        version: '~> 0.4'
-        condition: "Gem::Requirement.create('< 6.11.0').satisfied_by?(Gem::Version.new(puppet_version.dup))"
         platforms:
           - mswin
           - mingw


scotje:
Still red due to PDK/bundler bug that I'm working on.

sanfrancrisko:
I think this may be causing an issue on the Cloud CI pipelines @DavidS , @carabasdaniel , @sheenaajay . Hopefully my understanding is correct, but it seems like Litmus is not being added as a dependency due to the new condition added in this PR. If I specify a version in `PUPPET_GEM_VERSION`, then perform a `bundle install`, I do get Litmus, however, on the CI pipeline, we depend on the [`matrix_from_metadata script`](https://github.com/puppetlabs/puppet_litmus/blob/main/exe/matrix_from_metadata#L38-L42) regardless as to whether this ENV VAR is set.

Not sure what the best way to proceed is?

sanfrancrisko:
Chatting with @carabasdaniel , for now, I'll redo the PDK Update with an earlier template. After a conversation with @carabasdaniel  and @tphoney , we may want to [move Litmus from the system -> dev PMG](https://github.com/puppetlabs/puppet-module-gems/blob/cb32c6af5aff3899d628a2fe6958aadd6ca6ebae/config/dependencies.yml#L106-L117). According to @tphoney it has historically been in `system` to avoid clashes with `beaker`

DavidS:
> I think this may be causing an issue on the Cloud CI pipelines @DavidS , @carabasdaniel , @sheenaajay . Hopefully my understanding is correct, but it seems like Litmus is not being added as a dependency due to the new condition added in this PR. If I specify a version in `PUPPET_GEM_VERSION`, then perform a `bundle install`, I do get Litmus, however, on the CI pipeline, we depend on the [`matrix_from_metadata script`](https://github.com/puppetlabs/puppet_litmus/blob/main/exe/matrix_from_metadata#L38-L42) regardless as to whether this ENV VAR is set.

This seems entirely plausible.

> Chatting with @carabasdaniel , for now, I'll redo the PDK Update with an earlier template. After a conversation with @carabasdaniel and @tphoney , we may want to [move Litmus from the system -> dev PMG](https://github.com/puppetlabs/puppet-module-gems/blob/cb32c6af5aff3899d628a2fe6958aadd6ca6ebae/config/dependencies.yml#L106-L117). According to @tphoney it has historically been in `system` to avoid clashes with `beaker`

recent changes to the puppet-module-gems replace beaker with litmus in the system_tests group to remove beaker from our sphere of responsibility.

> Not sure what the best way to proceed is?

* revert this change as it is replacing the original problem (litmus conflicts puppet<6.x) with a different problem: litmus not available in some situations
* enhance this fix to guard against the case where PUPPET_GEM_VERSION is not set. In this case bundler would always install the latest puppet version, so litmus can be installed alongside it with no problem
* maybe identify situations where PUPPET_GEM_VERSION is set, and ensure that `--without system_tests` is also set to avoid the necessity for this guard at all

PR #387, opened 2021-01-11
Allow passing even if tear-down fails in Github Action CI

carabasdaniel:


DavidS (approved):
* did correctly form a github action workflow
* errors not related to this change
* seems to be semantically correct (didn't test the actual implementation)

PR #391, opened 2021-01-19
system_tests group handling

DavidS:
## Problem Description

Since moving `puppet_litmus` to the `system_tests` group, a few issues have cropped up:

* on older ruby versions or when running older puppet versions, litmus is not a valid install, through the litmus->bolt->puppet>=6 dependency chain. This happens:
  * in some unit-test cases we need to run older puppet versions
  * in the PDK when choosing older puppet versions 

Initial attempts to address this included avoiding installing litmus conditional on the `PUPPET_GEM_VERSION`. This turned out to be more complicated based on the limits of the `Gem::Requirement` API used and the diverse set of use-cases of the `PUPPET_GEM_VERSION`: 
* PDK provides a single-digit major puppet version
* unit testing in CI commonly provides optimistic version constraints like `~> 5.0`

The latter form does not allow comparison in the gem API, rendering this approach untenable. See #385, #386 and #389 for reference.

In https://github.com/puppetlabs/pdk/pull/935 I've even implemented forcing the PDK to make bundler ignore the `system_tests` group in regular operations.

## This Change

This PR reverts the original three PRs and instead makes the `system_tests` group optional by using bundler's `optional` flag to do so.

* this ensures that only environments that explicitly request litmus (through use of `BUNDLE_WITH=system_tests`) will have litmus installed
* with litmus' experimental state that gives us another layer of protection from accidents in customer environments
* this way we don't need a PDK change, which reduces time to repair by a lot
* this change also fixes up the CI configuration files to set BUNDLE_WITH where required and remove now redundant `--without system_tests` 

## Open Issues

* This requires `BUNDLE_WITH=system_tests` set in all environments - especially IC's local dev environments - where litmus is to be used
* **MORE PROBLEMATIC**: This makes litmus unavailable in the PDK, as `BUNDLE_WITH` does not get passed through to the `pdk bundle` subcommand:
```
$ env | sort | grep WITH
BUNDLE_WITH=system_tests
david@zion:~/git/pdksync/modules_pdksync/puppetlabs-powershell (pdksync)$ pdk bundle exec env | sort | grep WITH
pdk (INFO): Using Ruby 2.5.8
pdk (INFO): Using Puppet 7.1.0
$ 
```

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -589,9 +589,20 @@ Gemfile:
         version: '~> 1.0'
         condition: "Gem::Requirement.create(['>= 6.11.0', '< 8.0.0']).satisfied_by?(Gem::Version.new(puppet_version.dup))"
         platforms: ruby
+      - gem: 'puppet-module-posix-system-r#{minor_version}'
+        version: '~> 0.4'
+        condition: "Gem::Requirement.create('< 6.11.0').satisfied_by?(Gem::Version.new(puppet_version.dup))"
+        platforms: ruby
       - gem: 'puppet-module-win-system-r#{minor_version}'
         version: '~> 1.0'
         condition: "Gem::Requirement.create(['>= 6.11.0', '< 8.0.0']).satisfied_by?(Gem::Version.new(puppet_version.dup))"
+        platforms:
+          - mswin
+          - mingw
+          - x64_mingw
+      - gem: 'puppet-module-win-system-r#{minor_version}'
+        version: '~> 0.4'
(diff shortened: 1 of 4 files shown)


scotje:
It's easy enough to tweak the allow list for environment variables, specific to `pdk bundle` if we want. That will only help future PDK releases of course...

We could also introduce a flag/env-var for PDK to opt-in to Litmus/system testing. (E.g. `--with-litmus` or `--with-system-testing`) which sets `BUNDLE_WITH` under the hood. That would allow us to fiddle with the implementation down the line without changing the user-facing API.

PR #392, opened 2021-01-21
Revert litmus protection

DavidS:
This has passed on https://github.com/puppetlabs/puppetlabs-powershell/pull/328 

I believe that the existing protections are enough for anything except `pdk --puppet-version=` and explicit CI setups with versions between 6.0 and 6.11, which are not versions that we support in [PE versions](https://puppet.com/docs/pe/2019.8/component_versions_in_recent_pe_releases.html).

Closes #391 , which is a different, more complex approach to this, which would also require changes to the PDK.

This also obsoletes https://github.com/puppetlabs/pdk/pull/935 for now.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -588,11 +588,11 @@ Gemfile:
     ':system_tests':
       - gem: 'puppet-module-posix-system-r#{minor_version}'
         version: '~> 1.0'
-        condition: "Gem::Requirement.create(['>= 6.11.0', '< 8.0.0']).satisfied_by?(Gem::Version.new(puppet_version.dup)) || puppet_version.nil?"
+        condition: "Gem::Requirement.create(['>= 6.11.0', '< 8.0.0']).satisfied_by?(Gem::Version.new(puppet_version.dup))"
         platforms: ruby
       - gem: 'puppet-module-win-system-r#{minor_version}'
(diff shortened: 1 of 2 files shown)


scotje:
I think the reason I added these was that packaging breaks for the incompatible puppet versions. (During packaging, we run `bundle install` for every `--puppet-version` we ship with to ensure that all dependencies are available.)

DavidS:
Still trying to get passing tests on https://github.com/DavidS/pdk-vanagon/tree/test-revert-litmus-protection

DavidS:
Merging this for now to unblock IAC team's deploying of unrelated changes. If the other PDK changes around this were not enough to completely remedy the litmus-in-pdk situation, I'm happy to revisit once the pipelines are repaired (see [my comment here](https://tickets.puppetlabs.com/browse/PDK-1682?focusedCommentId=780440&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-780440)) when we can see if it is still an issue.

scotje:
OK, so the issue now seems to be that we still are skipping installing system test gems on the additional ruby versions (https://github.com/puppetlabs/pdk-vanagon/pull/271) so now there are missing gems in the packages. I'll try reverting that change and see if we can still build packages.

PR #399, opened 2021-02-13
Make litmus[variables] optional

cdenneen:
Fix issue where litmus variables was necessary and should have been optional.


DavidS (approved):
Thanks!

PR #400, opened 2021-02-14
Needed before script to be copied into the before_script for litmus

cdenneen:
Default before_script needs to be copied into the litmus test in order to do the `bundle install`


DavidS (approved):
Thanks for the fix!

PR #404, opened 2021-02-16
Use rspec-mocks as default mocking library

DavidS:
Previously puppetlabs_spec_helper was loading both rspec-mocks and mocha.
This leads to confusing inconsistencies in unit tests. For the longest time
we've been recommending people to default to rspec-mocks. This commit
makes it official for PDK2.0.

If your test suite is not yet ready for this step, remove the default value
through setting `mock_with: ~` (YAML's nil value) to get the PDK1.x behavior.


cdenneen:
@DavidS please add the `mock_with: ~` syntax to docs?

DavidS:
@cdenneen good call. Updated the README.

michaeltlombardi (approved):

PR #408, opened 2021-02-19
Add manual module publish and tag workflow using FORGE_API_KEY secret

carabasdaniel:


carabasdaniel:
Example run: https://github.com/carabasdaniel/puppetlabs-testing/actions/runs/617506533

moduleroot/.github/workflows/release.yml.erb
```diff
@@ -0,0 +1,57 @@
+<% common = config_for('common') -%>
+name: "Publish module"
+
+on:
+  workflow_dispatch:
+    inputs:
+      forge_api_key:
+        description: "Forge API token to upload module to the Puppet Forge"
```
DavidS, line 8: can we set this as github secret instead?

moduleroot/.github/workflows/release.yml.erb
```diff
@@ -0,0 +1,57 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.ref }}
+          clean: true
+      - name: "PDK Build"
+        uses: docker://puppet/pdk:nightly
+        with:
+          args: 'build'
+      - name: "Set token"
```
DavidS, line 51:
```suggestion
      - name: "mask token"
```

DavidS (commented):
one question inline

moduleroot/.github/workflows/release.yml.erb
carabasdaniel: Initial approach was to have it as a Github secret, but having it as an input might be easier for the future.
I'm open to suggestions here.

DavidS: If you want to preserve it as input for testing/one-off releases, we could just do both and use `${FORGE_TOKEN:-${{ env.FORGE_TOKEN }}}` (see "Use Default Values" on https://linux.die.net/man/1/bash) at the call site

carabasdaniel: Sounds like a good idea, I'll add that in

carabasdaniel: Having issues masking the input token, so for initial workflow we'll use only the repository secret

adrianiurca (approved):

PR #417, opened 2021-03-15
Add static & syntax checks to spec workflow

sanfrancrisko:
Adds the static and syntax checks to the `config_defauly.yml` for
`.github/workflows/spec.yml`

Adds an additional step to the `Spec` workflow to run these checks.

Also corrects a typo in a previous step.

See https://github.com/puppetlabs/puppetlabs-testing/pull/354 for test output

Closes #417

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -640,3 +640,6 @@ spec/spec_helper.rb:
   unmanaged: true
 .github/workflows/auto_release.yml:
   unmanaged: true
+.github/workflows/spec.yml:
+  unmanaged: true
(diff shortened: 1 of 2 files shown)


config_defaults.yml
```diff
@@ -640,3 +640,5 @@ spec/spec_helper.rb:
   unmanaged: true
 .github/workflows/auto_release.yml:
   unmanaged: true
+.github/workflows/spec.yml:
```
DavidS, line 643: shouldn't this be `unmanaged:true`, too?
```suggestion
.github/workflows/spec.yml:
  unmanaged: true
```

carabasdaniel (dismissed):
👍

carabasdaniel (approved):

PR #429, opened 2021-04-16
Misc cleanup

jeffbyrnes:
* Clean up per YAMLlint
* Clean up per RuboCop
* Use `bash` for `.travis/install_pdk.sh` b/c POSIX `sh` does not include `local`, but `bash` does


sanfrancrisko (approved):
Thanks for the contribution @jeffbyrnes

PR #431, opened 2021-04-26
Cleanup backtrace in spec tests

DavidS:
This needs thorough review to make sure the patterns match what is not necessary in a PDK installed unit test.

### DIFF
--- a/moduleroot/spec/spec_helper.rb.erb
+++ b/moduleroot/spec/spec_helper.rb.erb
@@ -62,6 +62,18 @@ RSpec.configure do |c|
     RSpec::Puppet::Coverage.report!(<%= @configs['minimum_code_coverage_percentage'] %>)
     <%- end -%>
   end
+
+  # Filter backtrace noise
+  backtrace_exclusion_patterns = [
+    /spec_helper/,
+    /gems/
+  ]
+
+  if c.respond_to?(:backtrace_exclusion_patterns)
+    c.backtrace_exclusion_patterns = backtrace_exclusion_patterns
+  elsif c.respond_to?(:backtrace_clean_patterns)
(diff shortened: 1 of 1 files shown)


DavidS:
Big thanks https://github.com/simp/pupmod-simp-simp/blob/f21c36180749c7998e091b8340d96e299eb45b1f/spec/spec_helper.rb#L107-L117 for the inspiration

rostetter:
This works for me!  I'd love to see it accepted.

I was able to put this in my spec/spec_helper_local.rb file as follows:

default_fixtures_path = File.expand_path(File.join(File.dirname(__FILE__), 'fixtures'))

RSpec.configure do |c|

  # Useless backtrace noise
  backtrace_exclusion_patterns = [
    /spec_helper/,
    /gems/
  ]

  if c.respond_to?(:backtrace_exclusion_patterns)
    c.backtrace_exclusion_patterns = backtrace_exclusion_patterns
  elsif c.respond_to?(:backtrace_clean_patterns)
    c.backtrace_clean_patterns = backtrace_exclusion_patterns
  end

  c.after(:suite) do
    RSpec::Puppet::Coverage.report!
  end
  # Hiera config file for unit tests
  c.hiera_config = File.join(default_fixtures_path, 'hiera/hiera.yaml')
end

Now my new, inexperienced puppet worker (and the rest of us) can more easily find the  actual error messages when we run pdk unit test without having to search through hundreds of lines of backtrace info which doesn't usually provide any value.

rostetter:
To make this pass the rubocop checks, we should change the regex block as follows:

  backtrace_exclusion_patterns = [
    %r{spec_helper},
    %r{gems},
  ]

sanfrancrisko (approved):

PR #432, opened 2021-04-29
Fix Bug in .rubocop.yml.erb

cocker-cc:
"enabled_cops = cop_configs.key" is nil, if Hash is empty, so catch this case in an If-Clause.


sanfrancrisko (approved):
Thanks for the fix @cocker-cc 👍

PR #441, opened 2021-06-22
Run validation steps prior to the matrix build

ekohl:
puppetlabs_spec_helper provides a task called validate which runs all static checks (syntax, lint, etc). If these fail, there isn't really any reason why you would want to start a full matrix build. Similarly, it also runs Rubocop prior to the matrix. Otherwise you just get a full failed matrix. The end result is less wasted CI on invalid syntax or style.

This is taken from Voxpupuli's modulesync_config.

I know this is incomplete and personally I never use PDK, but it came up in https://github.com/puppetlabs/puppetlabs_spec_helper/pull/340.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -87,7 +87,7 @@ common:
     script: bundle exec rake beaker
     stage: acceptance
   includes:
-    - env: CHECK="check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop syntax lint metadata_lint"
+    - env: CHECK="validate lint check rubocop"
       stage: static
     - env: PUPPET_GEM_VERSION="~> 6.0" CHECK=parallel_spec
       rvm: 2.5.7
@@ -114,7 +114,7 @@ appveyor.yml:
   use_litmus: false
   matrix:
     - RUBY_VERSION: 25-x64
-      CHECK: "syntax lint metadata_lint check:symlinks check:git_ignore check:dot_underscore check:test_file rubocop"
+      CHECK: "validate lint check rubocop"
     - PUPPET_GEM_VERSION: ~> 6.0
       RUBY_VERSION: 25
       CHECK: parallel_spec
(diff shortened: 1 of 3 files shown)


sanfrancrisko:
Would be keen to hear from @puppetlabs/modules as these changes could directly affect their workflows.

sanfrancrisko:
Change looks fine @ekohl - I was initially concerned that invoking the `validate` rake task only was going to reduce the amount of checks, but it seems to give us parity.

Not sure dropping out `check:symlinks check:git_ignore check:dot_underscore` would though - I don't see that being invoked from the [`validate` Rake task](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/a3b247312d41934112323fca5c74af42c52a8153/lib/puppetlabs_spec_helper/rake_tasks.rb#L216-L230).

ekohl:
> Not sure dropping out `check:symlinks check:git_ignore check:dot_underscore` would though - I don't see that being invoked from the [`validate` Rake task](https://github.com/puppetlabs/puppetlabs_spec_helper/blob/a3b247312d41934112323fca5c74af42c52a8153/lib/puppetlabs_spec_helper/rake_tasks.rb#L216-L230). I think we should probably retain those checks?

That's why the `check` task is there. I introduced it as a common task in:

https://github.com/puppetlabs/puppetlabs_spec_helper/commit/9a19f23ffe50c2d301c6c07fae26cebf1b50c9d0#diff-d45384b99d96b1c5f478804241e78b055919a0117ffe65182c9b8efb1152d006

Note that this does add the `check:test_file` task as well, which previously wasn't there.

In voxpupuli-test we also enhance the [check task with `check:trailing_whitespace`](https://github.com/voxpupuli/voxpupuli-test/blob/f6c85fe54412248339cbb8ebe197c0b95dbb5697/lib/voxpupuli/test/rake.rb#L10-L24) which is a trick module authors can also do.

sanfrancrisko:
Cool, thanks @ekohl .

@puppetlabs/modules as far as I can see, this will be OK for you folks as you've got the latest `puppetlabs_spec_helper` rolled out across the supported modules

carabasdaniel:
This looks good @ekohl, any reason why this is still a draft ?

ekohl:
Mostly that I wasn't sure if this was the right thing and forgot to mark it as ready. Done so now.

da-ar (approved):

carabasdaniel:
Thanks @ekohl, merged 👍

PR #442, opened 2021-06-27
Set `skip_branch_with_pr` to true by default in appveyor.yml template

TraGicCode:
If the owner of a repository creates a branch, pushes it, and then creates a PR, appveyor will not build it because of the branches filter (which is currently main & release).  This used to work but i'm guessing appveyor made a breaking change and this stopped at some point.

### DIFF
--- a/moduleroot/appveyor.yml.erb
+++ b/moduleroot/appveyor.yml.erb
@@ -1,5 +1,6 @@
 ---
 version: 1.1.x.{build}
+skip_branch_with_pr: true
 branches:
 <% if ((@configs['branches'] || []) - (@configs['remove_branches'] || [])).any? -%>
   only:


sanfrancrisko:
Thanks for highlighting @TraGicCode - I've been struggling to find any literature or announcements from AppVeyor mentioning any change in behaviour.

Also, reading further in to the `skip_branch_with_pr` feature, it looks as though subsequent commits to branches with PRs open would not have a new build triggered? See [this conversation](https://help.appveyor.com/discussions/questions/18437-skip_branch_with_pr-setting-doesnt-work). I'm thinking that this may cause confusion to some if this were the case. We may need to consider that case - I'll check what the behaviour is.

TraGicCode:
Hey @sanfrancrisko ,

So here is what `skip_branch_with_pr ` means:

1. If a branch is pushed and there is no PR then build it in appveyor
2. If a branch has a PR opened for it, no longer build the branch when commits are pushed and instead build the PR branch github created under the covers. 

I hope this makes sense.  number 2 is essentially this branch that travis is build right now for this open github issue

image

sanfrancrisko:
Appreciate the explanation @TraGicCode - LGTM, merging 👍

sanfrancrisko (approved):

PR #443, opened 2021-07-21
Only auto release if the changelog is updated

jarretlavallee:
Prior to this PR, the auto release workflow would always result in a
commit and PR. This PR updates the logic to only make the PR if the
changelog is updated.

This reduces the noise from invalid releases during the release schedule by only making PR's when there are updates to the changelog. An example of this workflow run can be seen in https://github.com/puppetlabs/puppetlabs-puppet_metrics_collector/runs/3124526271?check_suite_focus=true


sanfrancrisko (approved):
Thanks @jarretlavallee - this LGTM.

@puppetlabs/modules I'll defer to yourselves for the final approval & merge though

Issue #327, opened 2020-04-16
enabled_cops: all fails with NoMethodError


russellshackleford:
When using the hardcore rubocop profile in .sync.yml, I get the following

```
pdk (FATAL): Failed to render template '.rubocop.yml.erb'
NoMethodError: undefined method `keys' for "all":String
```

.sync.yml:
```
.rubocop.yml:
  selected_profile: hardcore
```

I'm using `tags/1.17.0-0-gd3a4319` on Ubuntu 18.04 with a new module that was created with `pdk new module`, then .sync.yml was created, followed by running `pdk update`.

DavidS:
I could reproduce this issue with the current version of the pdk-templates and PDK 1.18.1:

```
david@zion:~/tmp$ pdk new module hardcore --skip-interview
pdk (INFO): Creating new module: hardcore
pdk (INFO): Using the default template-url and template-ref.
pdk (INFO): Module 'hardcore' generated at path '/home/david/tmp/hardcore'.
pdk (INFO): In your module directory, add classes with the 'pdk new class' command.
david@zion:~/tmp$ cd hardcore/
david@zion:~/tmp/hardcore$ vi .sync.yml
david@zion:~/tmp/hardcore$ pdk update --force
pdk (FATAL): Failed to render template '.rubocop.yml.erb'
NoMethodError: undefined method `keys' for "all":String
david@zion:~/tmp/hardcore$ pdk --version
1.18.1
david@zion:~/tmp/hardcore$ cat .sync.yml 
---
.rubocop.yml:
  selected_profile: hardcore

david@zion:~/tmp/hardcore$ 
```

russellshackleford:
The error displayed has changed slightly since the original report. Here is the output for `selected_profile: off` and `selected_profile: hardcore` (respectively):

```
$ pdk update
pdk (FATAL): Failed to render template '.rubocop.yml.erb'
NoMethodError: undefined method `[]' for nil:NilClass

$ pdk update
pdk (FATAL): Failed to render template '.rubocop.yml.erb'
NoMethodError: undefined method `&' for {}:Hash
```

The `cleanups_only` and `strict` profiles work.

russellshackleford:
[#L36](https://github.com/puppetlabs/pdk-templates/blob/51828b4067b5e6a85165ad2fefb8d2086e4d5be8/moduleroot/.rubocop.yml.erb#L36) Appears to be the problem.

PR #453, opened 2021-08-27
(GH-327) Fix rubocop "off" & "hardcore" profiles

russellshackleford:
- Quoting the off hash key changed its value.
- enabled_cops conditional is always wrong because it isn't set yet
- hardcore should imply the kitchen sink, hence pending cops, too
- enabled cops is an array, not a hash.


michaeltlombardi (approved):
LGTM!

PR #460, opened 2022-01-26
disable legacy facts

bastelfreak:
Those are deprecated since ages and Vox Pupuli has this puppet-lint plugin enabled since a long time. O think Puppet Inc should have the same defaults as the community.


config_defaults.yml
```diff
@@ -1,6 +1,6 @@
 ---
 common:
-  disable_legacy_facts: false
+  disable_legacy_facts:true
```
skoef, line 3: perhaps put the space back in between `:` and `true`

bastelfreak: ah yes. fixed it.

chelnak:
Hey - I'm closing this PR given that this setting is present in spec_helper.

https://github.com/puppetlabs/puppetlabs_spec_helper/blob/01886e8fb8c9470e74ffde39445732ce06209125/lib/puppetlabs_spec_helper/rake_tasks.rb#L156

PR #466, opened 2022-02-23
Update TargetrubyVersion for rubocop

MartyEwings:
Prior to this commit the main template tested ruby syntax against ruby version 2.4
Since in the lowest Support release of Puppet, Ruby version 2.5 is in use, this is no longer the sensible default.

Defaulting to 2.5 ensures that code is linted to the most compatible ruby version for supported Puppet releases


bastelfreak (approved):

david22swan (approved):
LGTM

Issue #468, opened 2022-04-05
Enable puppet-lint-plugins also on Windows


ConradGroth:
## Use Case
I have a module that causes a puppet-lint warning on my build-server based on Linux, but not on my dev maschine based on Windows. I found out, that the warning actually came from one of the community plugins for puppet-lint.

## Describe the Solution You Would Like
Include puppet-lint-plugins on all platforms.

## Describe Alternatives You've Considered
Work in a VM or a Container.

## Additional Context
Removing the `platforms: ruby` from the Gemfile for the dependency `voxpupuli-puppet-lint-plugins` works fine on my Windows 10 maschine. By looking at some of the community plugins I don't see any reason why whis should be excluded on Windows. It's plain Ruby code.

jordanbreen28:
Completed in https://github.com/puppetlabs/pdk-templates/pull/469

PR #469, opened 2022-04-05
(packaging) Enable puppet-lint-plugins also on Windows #468

ConradGroth:


ConradGroth:
I tested this change locally on Windows 10 with PDK 2.4.0 (using Ruby 2.5.9 and Puppet 6.26.0). It runs perfectly and now I get the same warnings as under Linux.

ConradGroth:
@bastelfreak WDYT? The original addition of the puppet-lint-plugins with the restriction to platform ruby was made by you two months ago.

bastelfreak:
I think I did that because I had no Windows to test it.

david22swan (approved):
This change look's good to me and seems to run fine.
Going to go ahead and merge.
Thanks for putting the work in :)

PR #471, opened 2022-04-19
Disable Travis by default

binford2k:
Given that Travis no longer has a usable free tier and the recent
security vulnerabilities, I'd like to open the conversation about
disabling it by default.

Setting the file to `unmanaged` means that this will only affect new
modules--it will not delete existing configurations from modules
currently using it. Users can easily enable it by just setting the one
key and all the existing default configuration is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,13 @@ In this example the automated release prep workflow is triggered every Saturday 
 We can trigger automated builds with every change to our code base in the main branch, other branches or even a pull request.
 Travis uses a .travis.yml file in the root of your repository to learn about your project and how you want your builds to be executed.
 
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

chelnak:
I think this is a reasonable change. In addition to the reason in the description, it also give users one less file a to worry about when bootstrapping a new module.

smortex (approved):

logicminds:
Yes please unmanage that file.  Tired of always having to put it in my .sync file.   Should also unmanage appveyor as well.

david22swan:
@binford2k Is this good to merge?
The change look's good to me

chelnak:
Hey, i'm closing this as Travis has now been removed.

PR #481, opened 2022-10-04
(CONT-19) Removal of puppet-module-gems

david22swan:
With the recent updates to bundler the need for a specific set off gems for each version of ruby that we support has been removed.
As such the logic behind our gems is now much simpler and can easily fit within the default templates, thus removing our need to have a separate module to manage them, with the templates already managing all other similar default requirements for our modules.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -557,32 +557,39 @@ Gemfile:
   required:
     ':development':
       - gem: 'json'
+        version: '~> 2.0'
       - gem: 'voxpupuli-puppet-lint-plugins'
-        version: '>= 3.0.0'
+        version: '~> 3.0'
       - gem: 'facterdb'
-        version: '< 2.0.0'
+        version: '~> 1.18'
       - gem: 'metadata-json-lint'
-        version: '< 4.0.0'
-      - gem: 'puppet-lint'
-        version: '< 3.0.0'
+        version:
+        - '>= 2.0.2'
+        - '< 4.0.0'
       - gem: 'puppetlabs_spec_helper'
-        version: '< 4.0.0'
-      - gem: 'rspec-puppet'
-        version: '< 3.0.0'
+        version:
+        - '>= 3.0.0'
+        - '< 5.0.0'
       - gem: 'rspec-puppet-facts'
-        version: '< 3.0.0'
+        version: '~> 2.0'
       - gem: 'codecov'
+        version: '~> 0.2'
       - gem: 'dependency_checker'
+        version: '~> 0.2'
       - gem: 'parallel_tests'
+        version: '~> 3.4'
       - gem: 'pry'
-      - gem: 'serverspec'
+        version: '~> 0.10'
       - gem: 'simplecov-console'
-      - gem: 'specinfra'
+        version: '~> 0.5'
       - gem: 'puppet-debugger'
-      - gem: 'ed25519'
-        platforms: ruby
-      - gem: 'bcrypt_pbkdf'
-        platforms: ruby
+        version: '~> 1.0'
+      - gem: 'rubocop'
+        version: '= 1.6.1'
+      - gem: 'rubocop-performance'
+        version: '= 1.9.1'
+      - gem: 'rubocop-rspec'
+        version: '= 2.0.1'
       - gem: 'rb-readline'
         version: '= 0.5.5'
         platforms:
@@ -593,6 +600,8 @@ Gemfile:
       - gem: 'puppet_litmus'
         version: '< 1.0.0'
         platforms: ruby
+      - gem: 'serverspec'
+        version: '~> 2.41'
 .gitlab-ci.yml:
   defaults:
     cache:
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -556,46 +556,43 @@ Rakefile:
 Gemfile:
   required:
     ':development':
-      # hardcode JSON version to what's shipped in pdk for now
-      - gem: json
-        version: '= 2.0.4'
-        condition: "Gem::Requirement.create('~> 2.4.2').satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))"
-      - gem: json
-        version: '= 2.1.0'
-        condition: "Gem::Requirement.create(['>= 2.5.0', '< 2.7.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))"
-      - gem: json
-        version: '= 2.3.0'
-        condition: "Gem::Requirement.create(['>= 2.7.0', '< 2.8.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))"
-      - gem: 'puppet-module-posix-default-r#{minor_version}'
-        version: '~> 1.0'
-        platforms: ruby
-      - gem: 'puppet-module-posix-dev-r#{minor_version}'
-        version: '~> 1.0'
-        platforms: ruby
-      - gem: 'puppet-module-win-default-r#{minor_version}'
-        version: '~> 1.0'
-        platforms:
-          - mswin
-          - mingw
-          - x64_mingw
-      - gem: 'puppet-module-win-dev-r#{minor_version}'
-        version: '~> 1.0'
-        platforms:
-          - mswin
-          - mingw
-          - x64_mingw
+      - gem: 'json'
       - gem: 'voxpupuli-puppet-lint-plugins'
-        version: '>= 3.0'
-    ':system_tests':
-      - gem: 'puppet-module-posix-system-r#{minor_version}'
-        version: '~> 1.0'
+        version: '>= 3.0.0'
+      - gem: 'facterdb'
+        version: '< 2.0.0'
+      - gem: 'metadata-json-lint'
+        version: '< 4.0.0'
+      - gem: 'puppet-lint'
+        version: '< 3.0.0'
+      - gem: 'puppetlabs_spec_helper'
+        version: '< 4.0.0'
+      - gem: 'rspec-puppet'
+        version: '< 3.0.0'
+      - gem: 'rspec-puppet-facts'
+        version: '< 3.0.0'
+      - gem: 'codecov'
+      - gem: 'dependency_checker'
+      - gem: 'parallel_tests'
+      - gem: 'pry'
+      - gem: 'serverspec'
+      - gem: 'simplecov-console'
+      - gem: 'specinfra'
+      - gem: 'puppet-debugger'
+      - gem: 'ed25519'
         platforms: ruby
-      - gem: 'puppet-module-win-system-r#{minor_version}'
-        version: '~> 1.0'
+      - gem: 'bcrypt_pbkdf'
+        platforms: ruby
+      - gem: 'rb-readline'
+        version: '= 0.5.5'
         platforms:
           - mswin
           - mingw
           - x64_mingw
+    ':system_tests':
+      - gem: 'puppet_litmus'
+        version: '< 1.0.0'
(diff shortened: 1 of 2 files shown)


config_defaults.yml
```diff
@@ -556,46 +556,43 @@ Rakefile:
...
-      - gem: 'puppet-module-posix-default-r#{minor_version}'
-        version: '~> 1.0'
+      - gem: 'json'
+      - gem: 'voxpupuli-puppet-lint-plugins'
+        version: '>= 3.0.0'
+      - gem: 'facterdb'
+        version: '< 2.0.0'
+      - gem: 'metadata-json-lint'
+        version: '< 4.0.0'
+      - gem: 'puppet-lint'
```
ekohl, line 566: Isn't this redundant since `voxpupuli-puppet-lint-plugins' already depends on it:
https://github.com/voxpupuli/voxpupuli-puppet-lint-plugins/blob/56aa3a801627210093203ceeb1e309f291763e7b/voxpupuli-puppet-lint-plugins.gemspec#L17

config_defaults.yml
```diff
@@ -556,46 +556,43 @@ Rakefile:
...
-      - gem: json
-        version: '= 2.3.0'
-        condition: "Gem::Requirement.create(['>= 2.7.0', '< 2.8.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))"
-      - gem: 'puppet-module-posix-default-r#{minor_version}'
-        version: '~> 1.0'
+      - gem: 'json'
+      - gem: 'voxpupuli-puppet-lint-plugins'
+        version: '>= 3.0.0'
+      - gem: 'facterdb'
+        version: '< 2.0.0'
```
ekohl, line 563: It'd strongly suggest a lower bound for this and other gems.
```suggestion
        version: '~> 2.0'
```

config_defaults.yml
```diff
@@ -556,46 +556,43 @@ Rakefile:
...
+        version: '< 3.0.0'
+      - gem: 'rspec-puppet-facts'
+        version: '< 3.0.0'
+      - gem: 'codecov'
+      - gem: 'dependency_checker'
+      - gem: 'parallel_tests'
+      - gem: 'pry'
+      - gem: 'serverspec'
+      - gem: 'simplecov-console'
+      - gem: 'specinfra'
```
ekohl, line 580: This is only needed for system tests.

config_defaults.yml
```diff
@@ -556,46 +556,43 @@ Rakefile:
...
+        version: '< 4.0.0'
+      - gem: 'rspec-puppet'
+        version: '< 3.0.0'
+      - gem: 'rspec-puppet-facts'
+        version: '< 3.0.0'
+      - gem: 'codecov'
+      - gem: 'dependency_checker'
+      - gem: 'parallel_tests'
+      - gem: 'pry'
+      - gem: 'serverspec'
```
ekohl, line 578: This is only needed for system tests.

ekohl (commented):
I welcome this change.

chelnak:
> I welcome this change.

I'm really excited about this. 🥳

config_defaults.yml
david22swan: Hmm, it's been under the development namespace since puppet-module-gems was a thing?

ekohl: Well, then it was always wrong. You can even drop it if you want since serverspec requires it: https://github.com/mizzy/serverspec/blob/69e57466478ff4e0c248d3e86a75b3405643d44f/serverspec.gemspec#L24.

config_defaults.yml
```diff
@@ -556,46 +556,43 @@ Rakefile:
...
+      - gem: 'simplecov-console'
+      - gem: 'specinfra'
+      - gem: 'puppet-debugger'
+      - gem: 'ed25519'
         platforms: ruby
-      - gem: 'puppet-module-posix-dev-r#{minor_version}'
-        version: '~> 1.0'
+      - gem: 'bcrypt_pbkf'
         platforms: ruby
```
ekohl, line 585: `ed25519` and `bcrypt_pbkf` are needed by `net-ssh` but recent versions depend on them directly (see https://rubygems.org/gems/net-ssh). Really, they're only used for acceptance testing and can be dropped now.

config_defaults.yml
```diff
@@ -556,46 +556,43 @@ Rakefile:
...
+      - gem: 'facterdb'
+        version: '< 2.0.0'
+      - gem: 'metadata-json-lint'
+        version: '< 4.0.0'
+      - gem: 'puppet-lint'
+        version: '< 3.0.0'
+      - gem: 'puppetlabs_spec_helper'
+        version: '< 4.0.0'
+      - gem: 'rspec-puppet'
+        version: '< 3.0.0'
```
ekohl, line 571: `rspec-puppet` is also redundant IMHO: puppetlabs_spec_helper depends on it (https://github.com/puppetlabs/puppetlabs_spec_helper/blob/a28e6e9c761d186399577dcb86e80cfae3c302b6/puppetlabs_spec_helper.gemspec#L29).

config_defaults.yml
david22swan: Fair point

config_defaults.yml
david22swan: lower bounds added to everything

config_defaults.yml
david22swan: Moved

config_defaults.yml
david22swan: Removed

config_defaults.yml
```diff
@@ -557,37 +557,52 @@ Gemfile:
...
       - gem: 'json'
+        version: '~> 2.0'
       - gem: 'voxpupuli-puppet-lint-plugins'
-        version: '>= 3.0.0'
+        version: '~> 3.0'
       - gem: 'facterdb'
-        version: '< 2.0.0'
+        version:
+        - '>= 0.8.1'
+        - '< 2.0.0'
```
ekohl, line 566: Version 1.18.0 added Ubuntu 22.04 facts so I'd actually suggest:
```suggestion
        version: '~> 1.18'
```

config_defaults.yml
```diff
@@ -556,46 +556,66 @@ Rakefile:
...
+        version: '= 1.9.1'
+      - gem: 'rubocop-rspec'
+        version: '= 2.0.1'
+      - gem: 'ed25519'
+        version: '~> 1.2'
         platforms: ruby
-      - gem: 'puppet-module-posix-dev-r#{minor_version}'
+      - gem: 'bcrypt_pbkdf'
         version: '~> 1.0'
         platforms: ruby
```
ekohl, line 606: I'd suggest to drop both ed25519 and bcrypt_pbkdf since net-ssh already depends on them and you don't need them anywhere else.

config_defaults.yml
david22swan: Yeh, have checked and they've both been added as dependencies for at least 5 years

config_defaults.yml
david22swan: removed

config_defaults.yml
david22swan: Pinned

config_defaults.yml
david22swan: already done

moduleroot/Gemfile.erb
```diff
@@ -1,22 +1,38 @@
 <%
 def gem_length(gem)
-  if gem['from_env']
-    version_len = " *location_for(ENV['#{gem['from_env']}'])".length + ((" || '#{gem['version']}'".length if gem['version']) || 0)
+  if gem['version']
```
chelnak, line 3: There is some code duplication here with the function defined on line 21. I wonder if we could reduce this and share  some commonalities?

chelnak (changes requested):
This is amazing!!!

Just one small query from me.

moduleroot/Gemfile.erb
david22swan: Not without redoing a good bit of the template, the logic on how it works is a bit backward.

chelnak: Fair enough. Something that we can look at in the future then.

chelnak (dismissed):

config_defaults.yml
```diff
@@ -556,46 +556,54 @@ Rakefile:
...
+      - gem: 'facterdb'
+        version: '~> 1.18'
+      - gem: 'metadata-json-lint'
+        version:
+        - '>= 2.0.2'
+        - '< 4.0.0'
+      - gem: 'puppetlabs_spec_helper'
+        version:
+        - '>= 2.9.0'
+        - '< 4.0.0'
```
ekohl, line 572: 4.0.0 is out. Any reason not to allow it?

config_defaults.yml
```diff
@@ -556,46 +556,54 @@ Rakefile:
...
+        version: '~> 3.0'
+      - gem: 'facterdb'
+        version: '~> 1.18'
+      - gem: 'metadata-json-lint'
+        version:
+        - '>= 2.0.2'
+        - '< 4.0.0'
+      - gem: 'puppetlabs_spec_helper'
+        version:
+        - '>= 2.9.0'
```
ekohl, line 571: Why 2.9.0? Looking through https://github.com/puppetlabs/puppetlabs_spec_helper/blob/main/CHANGELOG.md I think you at least want 2.15.0 since that works with Ruby 2.7. Then it looks like you depend on the reconfigurability of puppet-lint in `Rakefile.erb` so that's at least 3.0.0.

config_defaults.yml
```diff
@@ -556,46 +556,54 @@ Rakefile:
...
+        - '< 4.0.0'
+      - gem: 'rspec-puppet-facts'
+        version:
+        - '>= 1.10.0'
```
ekohl, line 575: I'd strongly suggest `~> 2.0` for performance reasons.

ekohl (changes requested):

config_defaults.yml
david22swan: Same as above, was the original pin

config_defaults.yml
david22swan: Hmmm, cant see any reason not to raise the lower pin

config_defaults.yml
```diff
@@ -556,46 +556,52 @@ Rakefile:
...
+        version: '~> 0.2'
+      - gem: 'parallel_tests'
+        version: '~> 3.4'
+      - gem: 'pry'
+        version: '~> 0.10'
+      - gem: 'simplecov-console'
+        version: '~> 0.5'
+      - gem: 'puppet-debugger'
         version: '~> 1.0'
+      - gem: 'rubocop'
```
ekohl, line 587: As a hint: rubocop-rake also exists.

ekohl (approved):
:+1: on most of it. This code is now unused:
https://github.com/puppetlabs/pdk-templates/blob/510abaa4a35d4b357d20383cf45921195ba5d364/moduleroot/Gemfile.erb#L58-L59

chelnak (approved):

alexjfisher:
FWIW, users accessing rubygems via an Artifactory repository will want to read about this feature https://www.jfrog.com/confluence/display/JFROG/RubyGems+Repositories#RubyGemsRepositories-RetrievingLatestRubyGemsPackageCompatiblewithYourRubyVersions introduced in Artifactory 7.15.3

We're not running this version _yet_ and in recent months have had to modify our pdk-templates fork to pin more and more gems that were no longer supporting older rubies and no longer being pinned through updates to `puppet-module-gems`.

PR #484, opened 2022-11-30
voxpupuli-puppet-lint-plugins: Pull in 3.1 or newer

bastelfreak:
We released Version 3.1 today. It pulls in latest versions of the different plugins. This is basically a bugfix because we noticed that pdk doesn't automatically update available gems. With this simple change, the version in the Gemfile will be raised and pdk will ensure latest versions are installed, without the requirement of people running `pdk bundle update` (which could have sideeffects).


chelnak:
Thanks @bastelfreak.

chelnak (approved):

PR #488, opened 2023-01-14
(CONT-422) - Re-pin json gems

chelnak:
Json gems should be pinned to their default versions so that users aren't forced to recompile the native extensions.

See the following for a version matrix: https://stdgems.org/json/#gem-version


jordanbreen28 (approved):
Looks good

PR #490, opened 2023-01-26
Ensure that the devcontainer.json is valid JSON

zigsphere:
Currently, the devcontainer.json template file is invalid JSON resulting in a failed JSON parser error if people have a JSON parser in their CI/CD pipeline. Fixing this to avoid this issue moving forward when updating or creating a module with PDK.


chelnak (approved):
Hey, thanks!

PR #493, opened 2023-02-05
(docs) Adding example for Gemfile section

Tamerz:
I've added an example for updating the Gemfile since it wasn't completely obvious to me how it should look.


Tamerz:
I realized I missed that there is an example already but just without the two version constraints. I will close this request.

PR #494, opened 2023-02-10
(maint) Updates checkout action

mhashizume:
This commit updates all instances of the actions/checkout@v2 to actions/checkout@v3 in perparation for the former's deprecation as part of the NodeJS deprecation.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


chelnak (approved):
Good spot and thank you!

chelnak:
Ignoring the mend fail here - it wont work from forks atm. Something we are aware of and will fix.

PR #495, opened 2023-02-11
Update Gitpod Dockerfile

roock:
- Improve Docker Layering (remove temporary file in the same layer they are downloaded)
- workspace-full Container base image has been updated to Ubuntu focal, so update puppet release to focal too.
- Update to Puppet 7
  - Building fails with Puppet 6 as latest version of puppet-debugger cannot be installed anymore with it (see below)

```
ERROR:  Error installing puppet-debugger:
        The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.3.26. Try installing it with `gem install bundler -v 2.3.26` and then running the current command again
        bundler requires Ruby version >= 2.6.0. The current ruby version is 2.5.0.
```


chelnak:
Hey, I'm closing this as GitPod has now been removed.

Thank you!

PR #506, opened 2023-04-04
(CONT-844) - Update workflow templates

jordanbreen28:
Prior to this PR, pdk-templates included templates for workflows that became obselete with the deployment of reusable workflows to all our supported modules.

This PR updates the workflow templates and config_defaults.yml, so that the correct template is applied when running pdk.


moduleroot/.github/workflows/auto_release.yml.erb
```diff
@@ -8,95 +8,7 @@ on:
...
-      run: |
-        echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
-        echo "Pull Request URL - ${{ steps.cpr.outputs.pull-request-url }}"
-<% if common['honeycomb']['enabled'] -%> 
-    - name: "Honeycomb: Record finish step"
-      if: ${{ always() }}
-      run: |
-        buildevents step $TRACE_ID $STEP_ID $STEP_START 'Finished auto release workflow'
-<% end -%>
+  release_prep:
```
GSPatton, line 12: is the erb syntax right here? Not sure how it's supposed to be myself but looks like you'll need `<% end -%>` etc

GSPatton: right you are

GSPatton (approved):
looks good

PR #507, opened 2023-04-04
Remove puppet-blacksmith requirement

MartyEwings:
Prior to this commit puppet-blacksmith or a version of puppet-module-gems with this gem pulled into it was required to run any local PDK rake tasks, including litmus, this gem is not used by default anywhere and should not be part of the template

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -15,7 +15,6 @@ require 'bundler'
 require 'puppet_litmus/rake_tasks' if Bundler.rubygems.find_name('puppet_litmus').any?
 require 'puppetlabs_spec_helper/rake_tasks'
 require 'puppet-syntax/tasks/puppet-syntax'
-require 'puppet_blacksmith/rake_tasks' if Bundler.rubygems.find_name('puppet-blacksmith').any?
 require 'github_changelog_generator/task' if Bundler.rubygems.find_name('github_changelog_generator').any?
 require 'puppet-strings/tasks' if Bundler.rubygems.find_name('puppet-strings').any?
 <% if ! @configs['requires'].nil? -%>


PaulSD:
@GSPatton
Could you create a new tag with this commit so that `pdk update` will get it?

Without this commit, various pdk commands fail, and if I manually change our `Rakefile` to avoid those failures then our periodic `pdk update` keeps reverting it and breaking things again.

GSPatton:
Hi @PaulSD. Just so I can understand your request better, how are you using pdk-templates that requires a tag on this commit?

PaulSD:
I'm running `pdk update`, which automatically pulls pdk-templates using the tag listed here: https://github.com/puppetlabs/pdk/blob/main/lib/pdk/version.rb#L3

That causes problems because the current tag (`2.7.4`) does not include this commit.

Without this commit, if PDK was installed using the `pdk` RPM (which no longer includes the puppet-blacksmith gem) then this puppet-blacksmith reference causes various failures when running various PDK commands.

This commit fixes that problem, but `pdk update` doesn't get the fix yet because this commit is not included in the `2.7.4` tag that is pulled by `pdk update`.

My understanding is that in order to fix this so that `pdk update` gets this commit, a new tag needs to be created in this repo (eg `2.7.5`) and then the tag listed in https://github.com/puppetlabs/pdk/blob/main/lib/pdk/version.rb#L3 needs to be updated to use that new tag.

PaulSD:
Ping @GSPatton
This is still biting us.  Any chance of getting a new tag/release of pdk-templates published?

chelnak:
Hey @PaulSD, you can use --template-url and --template-ref to control where your templates are sourced.

Im more than happy to tag pdk-templates though.

Just be aware that we won't be updating PDK until after Puppet 8 work is complete so it's likely that you will need to use --template-ref to fetch the latest tag.

chelnak:
@PaulSD pdk-templates is now tagged at 2.7.5.

Have a nice weekend!

PaulSD:
Thank you!

PR #515, opened 2023-05-11
Allow puppet-lint fail_on_warnings to be disabled

### LINKED ISSUE #514: Allow puppet-lint fail_on_warnings to be explicitly disabled
https://github.com/puppetlabs/pdk-templates/blob/d05508f29015fed8099d9b2a38fe1851dc351825/moduleroot/Rakefile.erb#L69-L71

There doesn't seem to be a way to trivially disable puppet-lint fail_on_warnings in the `lint` rake task as the default is `true` via https://github.com/puppetlabs/puppetlabs_spec_helper/blob/01886e8fb8c9470e74ffde39445732ce06209125/lib/puppetlabs_spec_helper/rake_tasks.rb#L156 and the template only supports setting to `true`. Perhaps this fragment could be changed to something like this?
```ruby
<% if @configs.include?('linter_fail_on_warnings') -%>
<% if @configs['linter_fail_on_warnings'] -%> 
PuppetLint.configuration.fail_on_warnings = true 
<% else -%>
PuppetLint.configuration.fail_on_warnings = false
<% end -%>
<% end -%> 
```
This would also match the description of the setting:

<html>
<body>

linter_fail_on_warnings | A boolean indicating if the linter should exit non-zero on warnings as well as failures. Does affect .puppet-lint.rc.
-- | --

</body>
</html>

nabertrand:
Closes #514

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -66,8 +66,8 @@ end
 <% checks.each do |check| -%>
 PuppetLint.configuration.send('disable_<%= check %>')
 <% end -%>
-<% if @configs['linter_fail_on_warnings'] -%>
-PuppetLint.configuration.fail_on_warnings = true
+<% if ! @configs['linter_fail_on_warnings'].nil? -%>
+PuppetLint.configuration.fail_on_warnings = <%= @configs['linter_fail_on_warnings'] %>
 <% end -%>
 <% linter_options =  @configs['linter_options'] || [] -%>
 <% linter_options.each do |option| -%>


chelnak (approved):

moduleroot/Rakefile.erb
```diff
@@ -66,8 +66,8 @@ end
 <% checks.each do |check| -%>
 PuppetLint.configuration.send('disable_<%= check %>')
 <% end -%>
-<% if @configs['linter_fail_on_warnings'] -%>
-PuppetLint.configuration.fail_on_warnings = true
+<% if ! @configs['linter_fail_on_warnings'].nil? -%>
+PuppetLint.configuration.fail_on_warnings = <%= @configs['linter_fail_on_warnings'] %>
```
jordanbreen28, line 70: Hey @nabertrand! Apologies for the delay,
should we set a default value of true for`linter_fail_on_warnings` in `config_defaults.yml`?

moduleroot/Rakefile.erb
```diff
@@ -66,8 +66,8 @@ end
 <% checks.each do |check| -%>
 PuppetLint.configuration.send('disable_<%= check %>')
 <% end -%>
-<% if @configs['linter_fail_on_warnings'] -%>
-PuppetLint.configuration.fail_on_warnings = true
+<% if ! @configs['linter_fail_on_warnings'].nil? -%>
```
jordanbreen28, line 69:
```suggestion
<% if !@configs['linter_fail_on_warnings'].nil? -%>
```
small styling nick

moduleroot/Rakefile.erb
jordanbreen28: @nabertrand apologies. Missed this notification completely. 
Yeah okay, this makes sense. This is only added to the template if the user has a value set in their .sync.yml so my comments on this possibly being a breaking change don't actually apply, it will take their predetermined value.

Happy to merge!

moduleroot/Rakefile.erb
jordanbreen28: Inapplicable to this file

jordanbreen28 (approved):

PR #522, opened 2023-06-20
(CONT-1119) Add puppet-strings

chelnak:
## Summary
This PR adds puppet-strings to the default template Gemfile.

## Additional Context
Requires the following PR to be merged: https://github.com/puppetlabs/bolt/pull/3216

## Related Issues (if any)
PDK is unable to pull in newer versions of puppet-strings because of a dependency in bolt. We should also reference puppet-strings as a first party gem in module Gemfiles.

### DIFF
--- a/.github/workflows/labeller.yml
+++ b/.github/workflows/labeller.yml
@@ -1,0 +1,27 @@
+name: Labeller
+
+on:
+  issues:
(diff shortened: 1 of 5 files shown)


david22swan (approved):
LGTM

PR #540, opened 2023-11-23
(CAT-1606) - Adding gem group for release_pre

Ramesh7:
## Summary

We have multiple use cases where we are installing all gems every time which also lead to installing old gems or conflicts, so creating `release_prep` for releasing the module.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x] Manually verified.


david22swan (approved):
LGTM

PR #547, opened 2023-12-19
(MAINT) - Remove github_changelog gem footprint

Ramesh7:
## Summary

Cleanup activity - Removing the reference of github changelog.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/moduleroot/Rakefile.erb
+++ b/moduleroot/Rakefile.erb
@@ -15,7 +15,6 @@ require 'bundler'
 require 'puppet_litmus/rake_tasks' if Gem.loaded_specs.key? 'puppet_litmus'
 require 'puppetlabs_spec_helper/rake_tasks'
 require 'puppet-syntax/tasks/puppet-syntax'
-require 'github_changelog_generator/task' if Gem.loaded_specs.key? 'github_changelog_generator'
 require 'puppet-strings/tasks' if Gem.loaded_specs.key? 'puppet-strings'
 <% if ! @configs['requires'].nil? -%>
 <%   @configs['requires'].each do |item| -%>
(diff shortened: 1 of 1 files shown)


moduleroot/.rubocop.yml.erb
```diff
@@ -6,8 +6,9 @@ defaults = {
     'rubocop-rspec',
   ],
   'AllCops' => {
+    'NewCops' => 'enable',
     'DisplayCopNames' => true,
-    'TargetRubyVersion' => '2.6',
+    'TargetRubyVersion' => '2.7',
```
bastelfreak, line 11: Why is that changed? It doesn't relate to the PR title

Ramesh7: we manually change the ruby version on PDK update everytime, so thought to cover as part of this as its small and quick one.
Let me know if I need to cover as part of diff PR?

jordanbreen28: I think this needs to stay as ruby 2.6, to keep compatibility with the jruby versions still used in puppet server (IIRC)

Ramesh7: ahh i see, wasn't knowing that. Would like to know more about it. Thanks for pitching in @jordanbreen28.

Ramesh7: resolved, ready for review again.

moduleroot/.rubocop.yml.erb
```diff
@@ -6,6 +6,7 @@ defaults = {
     'rubocop-rspec',
   ],
   'AllCops' => {
+    'NewCops' => 'enable',
```
bastelfreak, line 9: I still think the rubocop config should not be changed here. The change isn't related to removing the changelog generator. Also I don't know if the current rubocop config handles all cops, enabling new cops could be a breaking change (that also depends on how you manage newer rubocop versions).

Ramesh7: done!!

LukasAud (approved):
LGTM

PR #549, opened 2024-01-15
(CAT-1684) Addition of CHANGELOG.md

david22swan:
## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/.github/workflows/release_prep.yml
+++ b/.github/workflows/release_prep.yml
@@ -6,10 +6,68 @@ on:
       version:
         description: "Template version, must match the version for the PDK release it is being bundled with."
         required: true
+        type: "string"
 
 jobs:
   release_prep:
-    uses: "puppetlabs/cat-github-actions/.github/workflows/module_release_prep.yml@main"
-    with:
-      version: "${{ github.event.inputs.version }}"
-    secrets: "inherit"
+    name: "Release prep"
+    runs-on: "ubuntu-20.04"
+
+    steps:
+
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 0
+
+      - name: "Generate changelog"
+        run: |
+          export GH_HOST=github.com
+          gh extension install chelnak/gh-changelog
+          gh changelog new --next-version v${{ github.event.inputs.version }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "setup ruby"
+        uses: "ruby/setup-ruby@v1"
+        with:
+          ruby-version: "3.2"
+          bundler-cache: "true"
+          bundler: 2.4.22
+
+      - name: "bundle environment"
+        run: |
+          echo ::group::bundler environment
+          bundle env
+          echo ::endgroup::
+
+      - name: "Check if a release is necessary"
+        id: "check"
+        run: |
+          git diff --quiet CHANGELOG.md && echo "release=false" >> $GITHUB_OUTPUT || echo "release=true" >> $GITHUB_OUTPUT
+
+      - name: "Commit changes"
+        if: ${{ steps.check.outputs.release == 'true' }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release_prep.yml
```diff
@@ -0,0 +1,15 @@
...
+      version:
+        description: "Module version to be released. Must be a valid semver string. (1.2.3)"
+        required: true
+
+jobs:
+  release_prep:
+    uses: "puppetlabs/cat-github-actions/.github/workflows/module_release_prep.yml@main"
+    with:
+      version: "${{ github.event.inputs.version }}"
+    secrets: "inherit"
```
jordanbreen28, line 15: new line missing

.github/workflows/release_prep.yml
```diff
@@ -0,0 +1,15 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Module version to be released. Must be a valid semver string. (1.2.3)"
+        required: true
+
+jobs:
+  release_prep:
+    uses: "puppetlabs/cat-github-actions/.github/workflows/module_release_prep.yml@main"
```
jordanbreen28, line 12: not actually sure pdk-templates can fit the reusable templates because of this step here https://github.com/puppetlabs/cat-github-actions/blob/6c374668d26258ac920cd6e11f481600b9bd2e76/.github/workflows/module_release_prep.yml#L35. Gem release prep has a similar step

jordanbreen28: if it doesn't fail on these steps we could use it i suppose

.github/workflows/release_prep.yml
```diff
@@ -0,0 +1,15 @@
+name: "Release Prep"
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Module version to be released. Must be a valid semver string. (1.2.3)"
```
jordanbreen28, line 7: should we update this you think

david22swan: Updated

.github/workflows/release_prep.yml
david22swan: Added

.github/workflows/release_prep.yml
david22swan: Ok yeh........
Think I'll need to just copy the reusable over and alter it.

.github/workflows/release_prep.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: "bundle environment"
+        run: |
+          echo ::group::bundler environment
+          bundle env
+          echo ::endgroup::
+
+      - name: "Update REFERENCE.md"
+        run: |
+          bundle exec rake strings:generate:reference
```
jordanbreen28, line 46: don't think this is needed

.github/workflows/release_prep.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: "Get version"
+        id: "get_version"
+        run: |
+          echo "version=$(jq --raw-output .version metadata.json)" >> $GITHUB_OUTPUT
```
jordanbreen28, line 51: this step is for the version of a module in metadata.json so can be removed too i'd say

david22swan: WOuldn't it also be passed to the changelog to set it there?

jordanbreen28: it gets its version from the input -> line 27 of this workflow

david22swan: removed
Wonder why its retrieved 2 different ways?

jordanbreen28: think that step retrieves the current version from a modules metadata.json so it can be updated later on in the module release workflow IIRC :) 

Anyways looks good - happy to merge

jordanbreen28 (approved):

PR #553, opened 2024-01-16
(CAT-1686) Upgrade Rubocop 1.48.1 to 1.50.0

LukasAud:
Following a recent team decision, we are implementing a Rubocop Upgrade, moving the version from 1.48.1 to 1.50.0. This should be the final version until Puppet 7 is unsupported. This change should help any module dependant on PDK to easily update.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,6 @@
 
 source 'https://rubygems.org'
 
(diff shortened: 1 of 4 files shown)


jordanbreen28 (approved):
Looks good.

PR #556, opened 2024-01-24
(CAT-1693) - Remove deprecated codecov gem

jordanbreen28:
## Summary

the codecov gem is deprecated, and no longer required in
puppetlabs_spec_helper. We should remove this from the pdk-templates
configu defaults, as it has stopped receiving updates and is restricting
us on the versions of simplecov we can consume.

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


gavindidrichsen (approved):
LGTM

PR #559, opened 2024-02-09
allow overriding default gems

h0tw1r3:
Useful for testing changes to default gems. I'm using this to test puppet_litmus changes.
Currently there is no way to override default gems other than forking the pdk-templates repo.


jordanbreen28 (approved):
@h0tw1r3 This is great!
Tested and verified!

PR #560, opened 2024-02-15
puppetlabs_spec_helper: Require 7.X

bastelfreak:
We're using this since some time at Vox Pupuli and I think Puppet should also use the latest version of their own software. Tested for development group in https://github.com/puppetlabs/puppetlabs-apt/blob/main/Gemfile#L26

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -543,7 +543,7 @@ Gemfile:
       - gem: 'puppet-strings'
         version: '~> 4.0'
       - gem: 'puppetlabs_spec_helper'
-        version: '~> 6.0'
+        version: '~> 7.0'
     ':system_tests':
       - gem: 'puppet_litmus'
         version: 


bastelfreak:
https://github.com/puppetlabs/puppetlabs-apt/actions/runs/7916102903/job/21609282764 and this uses psh 7 for a release.

bastelfreak:
Is that a public pipeline?

david22swan:
Ah, disregard the above comment sorry, ran a clean image and the failures have migrated there. Just unfortunate timing on their appearance it seems.

bastelfreak:
@david22swan is there any update on this? I need the newer puppetlabs_spec_helper for a few customers.

david22swan:
@bastelfreak Still working on failures on the PDK sorry.
The original ones I last mentioned have been resolved however new ones appeared following that.

bastelfreak:
What kind of errors? something we can help with?

david22swan:
@bastelfreak Sorry for the wait on this, but we've got the PDK back into working order so I am happy to merge.
I've already done some preliminary testing so if you can resolve the conflict that has appeared during the wait I can merge immeaditely.

bastelfreak:
@david22swan rebased it

david22swan (approved):
LGTM

PR #568, opened 2024-03-07
(CAT-434) Update README.md

LukasAud:
Previous to this commit, our PDK-template documentation would be fairly intimadating and confusing, often jumpin from basic introductions to advanced concepts, while expecting the reader to have an understanding of concepts they may not be familiar with.

This commit aims to rework the README to ensure that its more readable and user-friendly.


gavindidrichsen (dismissed):
Excellent re-write.  Very clear.

README.md
```diff
@@ -3,33 +3,155 @@
...
+
+Target files are created by rendering a corresponding template which refers its configuration via the `@configs` hash.
+
+You can configure PDK to make files executable if the file is executable in the template. To do this, include `manage_execute_permissions: true` either in the target file or globally via the common config key.
 
 ## Basic Usage
 
-Templates like this one can be used in conjunction with the PDK. As default the PDK itself uses the templates within this repository to render files for use within a module. Templates can be passed to the PDK as a flag for several of the commands.
+By default PDK uses the templates within this repository to render files for use within a module. To use a third-party template, pass its URL using the `–-template-url` flag as shown in the following example:
```
jordanbreen28, line 47: might be useful to include an example for using `--template-ref` as well.
something like `pdk update --template-url https://github.com/my-org/pdk-templates --template-ref my_branch`

README.md
```diff
@@ -58,12 +180,11 @@ The following is a description and explanation of each of the keys within config
...
+> These workflows are depending on puppet-internal resources and are currently not suited for public consumption. Feel free to take them as inspiration how to run some tests on Github Actions. Please let us know at <ia_content@puppet.com> what you come up with!
 
 ### .github/workflows/release_prep.yml
 
-The auto release workflows prepares a module release PR. By default the workflow can be triggered manually when a release preparation PR needs to be created, however it allows setting a cron based trigger that can run automatically.
-To set up the automated release cron you can add a configuration to your .sync.yml file that matches the following example:
+> The auto release workflows prepares a module release PR. By default the workflow can be triggered manually when a release preparation PR needs to be created, however it allows setting a cron based trigger that can run automatically. To set up the automated release cron you can add a configuration to your .sync.yml file that matches the following example:
```
jordanbreen28, line 187:
```suggestion
> The auto release workflows prepares a module release PR. By default the workflow can be triggered manually when a release preparation PR needs to be created, however it allows setting a cron based trigger that can run automatically. To set up the automated release cron you can add a configuration to your .sync.yml file that matches the following example:
```
The naming of the workflows here is outdated.
auto-release -> release_prep

README.md
```diff
@@ -3,33 +3,155 @@
...
+```
+
+### Manage Rubocop rules
+
+To disable or enable certain Rubocop rules, use the following structure:
+
+```yaml
+.rubocop.yml:
+  default_configs:
+    Rule/Name:
```
jordanbreen28, line 146:
```suggestion
    Style/Documentation:
```
Might be better to include an actual example.
something like the above

README.md
```diff
@@ -3,33 +3,155 @@
...
+      Enabled: true/false
```
jordanbreen28, line 147:
```suggestion
      Enabled: false
```
think is more clear

README.md
```diff
@@ -3,33 +3,155 @@
...
+    - env: PUPPET_GEM_VERSION="~> 4.0" CHECK=parallel_spec
+      rvm: 2.1.9
+```
+
+### Removing default configuration values
+
+Values can be removed from the data passed to the templates using the [knockout prefix](https://www.rubydoc.info/gems/puppet/DeepMerge) `---` in `.sync.yml`.
+
+To remove a value from an array, prefix the value `---`. For example, to remove
+`2.5.1` from the `ruby_versions` array in `.travis.yml`:
```
jordanbreen28, line 83: same as above

README.md
```diff
@@ -3,33 +3,155 @@
...
-## Values of config\_defaults <a name="values"></a>
+Values can be added to the data passed to the templates, thus allowing you to make changes such as testing against additional operating systems or adding new rubocop rules.
 
-The following is a description and explanation of each of the keys within config\_defaults. This will help clarify the default settings we choose to apply to pdk modules.
+To add a value to an array simply place it into the `.sync.yml` file as shown below, here is the structure to add an additional unit test run against Puppet 4:
+
+```yaml
+.travis.yml:
+  includes:
+    - env: PUPPET_GEM_VERSION="~> 4.0" CHECK=parallel_spec
```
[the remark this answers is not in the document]
LukasAud: Can add, although I dont think it makes much of a difference in this scenario. The example is equally valid as long as its not a currently active puppet version

README.md
```diff
@@ -3,33 +3,161 @@
...
+
+> While we provide a basic template it is likely that it will not match what you need exactly, as such we allow it to be altered or added to through the use of the `.sync.yml` file.
+
+Here we will provide some examples of how you can use `.sync.yml`. For a more comprehensive list of default settings, check the [PDK default config values](#pdk-default-config-values) section. 
+
+### Adding configuration values
+
+Values can be added to the data passed to the templates, thus allowing you to make changes such as testing against additional operating systems or adding new rubocop rules.
+
+To add a value to an array simply place it into the `.sync.yml` file as shown below, here is the structure to add an additional unit test run against Puppet 4:
```
jordanbreen28, line 75:
```suggestion
To add a value to an array simply place it into the `.sync.yml` file as shown below, here is the structure to add an additional unit test run against Puppet 6:
```
To match the new example.
Could we just set to `~> 8.0` here to keep it up to date?
`rvm` could be updated to 3.2.0 just to ensure its a correct example as well

README.md
```diff
@@ -3,33 +3,161 @@
...
 
+#### Table of contents
+
+* [Introduction](#pdk-templates)
+* [How templates are used by PDK](#how-templates-are-used-by-pdk)
 * [Basic usage](#basic-usage)
-* [Values of config\_defaults](#values)
-* [Making local changes to the Template](#making-local-changes-to-the-template)
-* [Further Notes](#notes)
+* [Adjusting templates via /sync.yml](#adjusting-templates-via-syncyml)
```
nabertrand, line 23:
```suggestion
* [Adjusting templates via .sync.yml](#adjusting-templates-via-syncyml)
```
You could also keep the slash if it's meant to show it's at the root of the template repo, e.g. `/.sync.yml`

LukasAud: That was probably a spelling mistake. Will amend to adhere to the standard throughout the document. Thanks for pointing this out.

README.md
LukasAud: Will amend the message and bump rvm. But I believe the point of the example was to showcase adding new unit test cases against something that was not likely to be there. For example, unsupported puppet versions. thats why it wasnt updated to begin with. I could be misinterpreting the original documentation tho.

README.md
```diff
@@ -3,33 +3,161 @@
...
+
+Target files are created by rendering a corresponding template which refers its configuration via the `@configs` hash.
+
+You can configure PDK to make files executable if the file is executable in the template. To do this, include `manage_execute_permissions: true` either in the target file or globally via the common config key.
 
 ## Basic Usage
 
-Templates like this one can be used in conjunction with the PDK. As default the PDK itself uses the templates within this repository to render files for use within a module. Templates can be passed to the PDK as a flag for several of the commands.
+By default PDK uses the templates within this repository to render files for use within a module. To use a third-party template, pass its URL using the `–-template-url` flag as shown in the following example:
```
nabertrand, line 47: Should we note that the PDK will look for a branch called `main` unless specified? The error provided if that branch doesn't exist isn't clear that the branch couldn't be found, e.g.
```
pdk (FATAL): Unable to find a valid template at https://github.com/org/repo.git#main
```

LukasAud: I feel like it should be expected that the command would default to main. But I guess a wee note won't hurt. Specially if there is people out there still using master in their naming conventions.

jordanbreen28 (approved):

PR #569, opened 2024-03-11
(CAT-1743) Address .sync.yml not overriding cops

LukasAud:
Prior to this commit, it was noticed that Rubocop rules that were set to false in .sync.yml would not be properly overriden if the cop already existed as a default enabled cop.

This commit aims to correct this unintended functionality and adds a test case.


gavindidrichsen (approved):
LGTM

PR #577, opened 2024-05-14
puppet-lint: fail on warnings

bastelfreak:
IMO community best practice is it to fail on warnings. Why would you ignore all puppet-lint messages?

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -60,6 +60,7 @@ common:
 .yardopts:
   markup: markdown
 Rakefile:
+  linter_fail_on_warnings: true
   changelog_version_tag_pattern: 'v%s'
   default_disabled_lint_checks:
     - 'relative'


gavindidrichsen:
Hi @bastelfreak .  We've been discussing this change and think, while this may be best practice, this could be very disruptive to not only our own CI but also other organisations.

As this field is documented in the pdk-template [here](https://github.com/puppetlabs/pdk-templates?tab=readme-ov-file#rakefile), we feel this is sufficiently highlighted.

nabertrand:
I just wanted to add that I think the default is currently true as set here: https://github.com/puppetlabs/puppetlabs_spec_helper/blob/main/lib/puppetlabs_spec_helper/rake_tasks.rb#L156. I had to open https://github.com/puppetlabs/pdk-templates/pull/515 to even be able to disable failing on warnings.

bastelfreak:
@gavindidrichsen I don't think this is a breaking change. It's enabled by default in puppetlabs_spec_helper. This change in pdk-templates just makes it more explicit?

PR #602, opened 2024-11-05
(CAT-2138) Replace direct `rubocop` pins with  `voxpupuli-test`

david22swan:
This will align our Rubocop versions with Vox on a larger basis and also allow us to bring in the helper functions contained within this gem.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


david22swan:
Ok, gonna have to put a pin in this.
We are unable to find a valid version of this Gem

We require 9.2.0 for the puppetlabs_spec_helper requirement
But 9.1.0 drops rspec-puppet 4 which we still require

PR #604, opened 2024-11-18
(CAT-2158) Upgrade rexml to address CVE-2024-49761

amitkarsale:
## Summary

See https://nvd.nist.gov/vuln/detail/CVE-2024-49761

> REXML is an XML toolkit for Ruby. The REXML gem before 3.3.9 has a ReDoS vulnerability when it parses an XML that has many digits between &# and x...; in a hex numeric character reference (&#x...;). This does not happen with Ruby 3.2 or later. Ruby 3.1 is the only affected maintained Ruby. The REXML gem 3.3.9 or later include the patch to fix the vulnerability.

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -568,8 +568,7 @@ Gemfile:
       # Temporary Pin
       - gem: 'rexml'
         version: 
-          - '>= 3.0.0'
-          - '< 3.2.7'
+          - '>= 3.3.9'
     ':development, :release_prep':
       - gem: 'puppet-strings'
         version: '~> 4.0'


config_defaults.yml
```diff
@@ -568,8 +568,7 @@ Gemfile:
       # Temporary Pin
       - gem: 'rexml'
         version: 
-          - '>= 3.0.0'
-          - '< 3.2.7'
+          - '>= 3.3.9'
```
bastelfreak, line 571: why is rexml even in here?

amitkarsale: Sorry, I didnt understood the context. Do you mean `rexml`  gem should not be present here?

bastelfreak: Yes. i was wondering why it's present at all.

amitkarsale: Not sure about its presence, but just updated the template based on your suggestion on upgrading `rexml` version for a `CVE--2024-49761` fix

david22swan: This pin was added to resolve issues on Windows, newer versions of rexml where causing failures.

bastelfreak: @david22swan do you know if the pin is still required? I don't have a windows system to test.

david22swan: Looking into this now

david22swan:
Closing in favor of https://github.com/puppetlabs/pdk-templates/pull/606
Pin has been removed

PR #613, opened 2025-03-27
Only pin to the puppetcore versions.

gavindidrichsen:
Since the `PUPPET_GEM_VERSION` and `FACTER_GEM_VERSION` are not always going to be version contraints (they can also be a git URL), then I'm removing this flexible check and pinning only to the latest puppetcore gems.  If flexible puppet and facter gemsources are need via https or git, then remove the PUPPET_FORGE_TOKEN and use PUPPET_GEM_VERSION and/or FACTER_GEM_VERSION instead.

## Checklist
- [x] Manually verified.

### DIFF
--- a/moduleroot/Gemfile.erb
+++ b/moduleroot/Gemfile.erb
@@ -112,8 +112,8 @@ hiera_version = ENV.fetch('HIERA_GEM_VERSION', nil)
 # If PUPPET_FORGE_TOKEN is set then use authenticated source for both puppet and facter, since facter is a transitive dependency of puppet
 # Otherwise, do as before and use location_for to fetch gems from the default source
 if !ENV['PUPPET_FORGE_TOKEN'].to_s.empty?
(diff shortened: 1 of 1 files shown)


gavindidrichsen:
Checked both the cases where `PUPPET_FORGE_TOKEN` is unset and also set.  For example the Gemfile.lock output when set is as expected:

```
➜  example git:(development) ✗ cat Gemfile.lock | grep -E "(remote:|specs:| puppet | facter | hiera )"    
  remote: https://rubygems-puppetcore.puppet.com/
  specs:
    facter (4.11.0)
    puppet (8.11.0-universal-darwin)
      facter (>= 4.3.0, < 5)
  remote: https://rubygems.org/
  specs:
      puppet (>= 6.18.0)
      facter (< 5.0.0)
      puppet (>= 6)
      puppet (>= 7, < 9)
      facter (< 5)
      puppet (>= 7, < 9)
  facter (~> 4.11)!
  puppet (~> 8.11)!
➜  example git:(development) ✗ 
```

danadoherty639 (approved):

PR #618, opened 2025-06-09
(CAT-2297) Update Rubocop template to allow TargetRubyVersion to be set

david22swan:
## Checklist
- [x] Manually verified.


david22swan:
Have verified change against puppetlabs-motd

gavindidrichsen (approved):

PR #510, opened 2023-04-24
(CONT-885) pdk-templates v3 prep

chelnak:
This PR prepares PDK templates for the next release.

Module dependencies have been updated and a number of features that are not used by puppetlabs have been removed.

PR #636, opened 2026-05-19
Remove '/..yml' from ignored files list

smiba:
Removes the bogus line "/..yml" from config_defaults.yml introduced in [428f39ccd](https://github.com/puppetlabs/pdk-templates/commit/428f39ccd) on https://github.com/puppetlabs/pdk-templates/pull/510

Found this during a code review of our module after running pdk convert, traced it back to this change. 

Originally this line used to say "/.rubocop.yml" before it was changed to "/..yml" a few years ago

### DIFF
--- a/config_defaults.yml
+++ b/config_defaults.yml
@@ -58,7 +58,6 @@ common:
     - '/Rakefile'
     - '/rakelib/'
     - '/.rspec'
-    - '/..yml'
     - '/.yardopts'
     - '/spec/'
     - '/.vscode/'


smiba:
@LukasAud Could I get some feedback on this? Should be a really straight forward PR, been trying to get attention to this mistake for months now and no one from puppet has replied if this is an intentional change or an accidental mistake as I presume it is

david22swan:
Closing in favour of: https://github.com/puppetlabs/pdk-templates/pull/650

As yu stated this looks to be a mistake and as such the better option would be to revert it back to `.rubocop.yml`.
Thank you for bringing this to our attention however.

PR #655, opened 2026-08-06
Update puppet-strings gem version to ~> 5.0

david22swan:
## Summary
Puppet strings v4 brings in a version of `yard` with several CVEs against it.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.


LukasAud (approved):
LGTM

PR #666, opened 2026-09-14
(maint) Correct minimum Puppet 9 Ruby version

mhashizume:
While puppet-agent 9.x ships with Ruby 4.0, Puppet's actual minimum Ruby version is 3.4. This is in order to match with the minimum Ruby version of puppetserver, which uses JRuby and has older Ruby requirements.

This commit correct's Puppet 9's minimum Ruby version from 4.0 to 3.4.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


mhashizume:
Here's where Puppet's minimum Ruby version is set in its gemspec: https://github.com/puppetlabs/puppet-private/blob/9.0.0/puppet.gemspec#L7

